PR #3, opened 2015-12-30
9.0 account operating unit

JordiBForgeFlow:
- Adds the Operating Unit in the account move line.
- Define an Inter-operating unit clearing account at company level.
- When users create a journal entry with lines in different operating units,
  at the time of posting the journal entry it automatically creates the corresponding
  lines in the Inter-operating unit clearing account, making each OU self-balancing.
- Introduces checks that prevent users from entering cross-operating unit
  journal entries using different accounts.


JordiBForgeFlow:
I'm closing this one, substituted by https://github.com/OCA/operating-unit/pull/6

PR #8, opened 2016-02-15
[8.0][ADD][account_voucher_operating_unit]

JordiBForgeFlow:
# 
# Account Voucher Operating Unit

This module introduces Operating Units to the Account Voucher model. It also
introduces security rules to manage access control only to users that can
operate on the OU of the voucher.
## Customer or Supplier Payments
- The Operating Unit that drives the Voucher is the one assigned to the
  default GL account of the Journal.
- The payment lines of the journal entries are assigned to this OU, and the
  Accounts Receivable or Accounts Payable lines that are created will belong
  to the OU of the AR / AP journal item that is being reconciled.
- Operating Unit A can thus pay invoices posted by Operating Units B and C.
## Customer or Supplier Receipts
- The Operating Unit is assigned to the voucher, and then to the journal
  items when the voucher is posted.
# Configuration

To configure this module, you need to:
- For each bank/cash GL account that you intend to use in customer or supplier
  payments you have to define the default Operating Unit that the payment
  will be posted to. Go to _Invoicing / Configuration / Accounts / Accounts_
  and search for your bank/cash accounts. Then add the default Operating Unit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+language: python
+sudo: false
+cache: pip
+
+python:
+  - "2.7"
+
+addons:
+  apt:
+    packages:
+      - expect-dev  # provides unbuffer utility
+      - python-lxml  # because pip installation is slow
+
+env:
+  global:
+  - VERSION="8.0" LINT_CHECK="1"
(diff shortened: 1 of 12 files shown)


JordiBForgeFlow:
@kittiu see this PR. Please notice the latest commit as well.

PR #9, opened 2016-07-12
[8.0] [ADD] [crm operating unit]

AaronHForgeFlow:
## CRM with Operating Units

- Adds the Operating Unit (OU) to leads and opportunities
- Adds rules to access leads and opportunities according user’s OU

### Usage
- Add the Operating Unit to a lead or an opportunity.
- Users will only see leads and opportunities belonging to their OU
- They will only create or modify leads or opportunities of their OU


AaronHForgeFlow:
prstalesmall

PR #10, opened 2016-07-13
[8.0][ADD] stock operating unit

AaronHForgeFlow:
## Stock with Operating Units

- Adds the Operating Unit (OU) to stock location, stock warehouse (WH), stock picking and stock move
- Adds rules to access stock records according user’s OU
- Adds retrictions to avoid creating stock picking where the company is different to the OU's company
- The location's OU and the WH's OU must match
- By default, the default OU of the user is added to the WH and the stock picking
- Stock moves must be related to a location (source or destination) that belongs to the requesting OU
- If an Internal Location is child to a parent Internal Location, parent must belong to the same Operating Unit.
### Installation

No dependencies apart from operating_unit and stock
### Configuration

No extra configuration needed.
### Usage

Assing an OU for a WH and to locations if needed
When Stock Move is associated to a Stock Picking, at least the Source or the Destination Location’s OUs must be the same as that of the Stock Picking.


AaronHForgeFlow:
not working on this anymore

PR #6, opened 2016-01-22
[9.0] [ADD] [account_operating_unit]

JordiBForgeFlow:
Substitutes https://github.com/OCA/operating-unit/pull/3

PR #18, opened 2016-08-19
[9.0][ADD] stock operating unit

AaronHForgeFlow:
# Stock with Operating Units

- Adds the Operating Unit (OU) to stock location, stock warehouse (WH), stock picking and stock move.
- Adds rules to access stock records according user’s OU.
- Adds retrictions to avoid creating stock picking where the company is different to the OU's company. The location's OU and the WH's OU must match.
- By default, the default OU of the user is added to the WH and the stock picking.
- Stock moves must be related to a location (source or destination) that belongs to the requesting OU.
  If an Internal Location is child to a parent Internal Location, parent must belong to the same Operating Unit.
## Installation

This module depends on the module account_operating_unit. That module allows the company to generate a balanced balance sheet by operating unit. The PR for that module is here: https://github.com/OCA/operating-unit/pull/6


AaronHForgeFlow:
cc @jbeficent

PR #19, opened 2016-08-22
[9.0][ADD] stock account with operating units

AaronHForgeFlow:
# Stock account with operating units

In the context of perpetual inventory (real time inventory valuation) this module creates journal entries when stock moves are posted between internal locations within the same company, but different OU’s.
### Configuration

The journal entries will be created only if you have defined the product to be managed under real-time inventory valuation.
### Usage

Create an Internal Transfer between two Locations that belong to different Operating Units. A journal entry will then be created, where the operating unit of the source location will be credited for the inventory valuation, at the product's cost, and the operating unit of the destination location will be debited.


AaronHForgeFlow:
cc @jbeficent

PR #21, opened 2016-08-22
[9.0][ADD] stock account operating unit

AaronHForgeFlow:
# Stock account with operating units

In the context of perpetual inventory (real time inventory valuation) this module creates journal entries when stock moves are posted between internal locations within the same company, but different OU’s.
### Configuration

The journal entries will be created only if you have defined the product to be managed under real-time inventory valuation.
### Usage

Create an Internal Transfer between two Locations that belong to different Operating Units. A journal entry will then be created, where the operating unit of the source location will be credited for the inventory valuation, at the product's cost, and the operating unit of the destination location will be debited.

### DIFF
--- a/stock_account_operating_unit/README.rst
+++ b/stock_account_operating_unit/README.rst
@@ -1,0 +1,78 @@
+.. image:: https://img.shields.io/badge/license-LGPLv3-blue.svg
+   :target: https://www.gnu.org/licenses/lgpl.html
+   :alt: License: LGPL-3
+
+=======================================
+Stock account moves with Operating Unit
+=======================================
+
+This module introduces the following features:
+- Creates account move lines when stock moves are posted between internal
+locations within the same company, but different OU’s.
+
+
+Configuration
+=============
+
+If your company is required to generate a balanced balance sheet by
+operating unit you can specify at company level that operating units should
+be self-balanced, and then indicate a self-balancing clearing account.
+
+#. Create an account for "Inter-OU Clearing" of type Regular.
+#. Go to *Settings / Companies / Configuration* and:
+
+   * Set the "Operating Units are self-balanced" checkbox.
+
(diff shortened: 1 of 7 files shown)


AaronHForgeFlow:
cc @jbeficent @darshan-serpent

stock_account_operating_unit/model/stock.py
```diff
@@ -0,0 +1,98 @@
...
+                    move.location_dest_id.company_id and
+                    move.operating_unit_id != move.operating_unit_dest_id
+            ):
+                err = False
+                if move.location_id.usage != 'internal' \
+                        and move.location_dest_id.usage == 'internal':
+                    err = True
+                if move.location_id.usage != 'internal' \
+                        and move.location_dest_id.usage == 'internal':
+                    err = True
```
guewen: The 2 checks above are identical, should one of them be `move.location_id.usage == 'internal' and move.location_dest_id.usage != 'internal'`?

stock_account_operating_unit/README.rst
```diff
@@ -26,7 +26,7 @@ be self-balanced, and then indicate a self-balancing clearing account.
 * Go to *Settings / Companies / Configuration* and:
    ** Set the "Operating Units are self-balanced" checkbox
    ** Set the "Inter-OU Clearing"  account in "Inter-operating unit clearing
-      account" field.
+   account" field.
```
LoisRForgeFlow: This was fine before the change. <a href="http://docutils.sourceforge.net/docs/user/rst/quickref.html#bullet-lists"> Continuing text must be aligned after the bullet and whitespace. </a>

AaronHForgeFlow: Yes I think the same, but travis does not agree

AaronHForgeFlow:
Can you please take a look @lreficent @darshan-serpent ?

AaronHForgeFlow:
Sorry, not working anymore on this.

PR #23, opened 2016-08-24
[9.0][ADD] sale stock operating unit

AaronHForgeFlow:
# Sale order lines  and stock moves with Operating Units
- A Delivery Order must have the same OU as that of the sales order that is associated to it.
- A Sales Order can only use a supplying Warehouse (shop in 7.0) that belongs to the same OU.
- A Delivery Order must have the same OU as that of the sales order that is associated to it.
### Usage
- No need to set the OU to the stock picking is created from a sales order


guewen (approved):

sale_stock_operating_unit/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Go to 'Settings / Technical / Actions / User-defined Defaults' and remove
+the default set for the Shop.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/213/7.0
```
serpentcs-dev1, line 26: 9.0

PR #26, opened 2016-09-15
[ADD][9.0] crm operating unit

AaronHForgeFlow:
- Adds the Operating Unit (OU) to the Lead.
- The user’s default Operating Unit (OU) is proposed at the time of creating the Lead / Opportunity.
- Security rules are defined to ensure that users can only see the Opportunity / Lead of that Operating Units in which they are allowed access to.


crm_operating_unit/models/crm_lead.py
```diff
@@ -0,0 +1,27 @@
...
+                                        self.env['res.users'].
+                                        operating_unit_default_get(self._uid))
+
+    @api.one
+    @api.constrains('operating_unit_id', 'company_id')
+    def _check_company_operating_unit(self):
+        if self.company_id and \
+                self.operating_unit_id and \
+                self.company_id != self.operating_unit_id.company_id:
+            raise Warning(_('Configuration error!\nThe Company in the\
```
elicoidal: Looks strange: maybe you should add a domain in the form view (show only OU from the CRM lead company) so this does not happen?

elicoidal (changes requested):
check the control on the company.
Non blocking but functionally dubious.
other than that :+1:

crm_operating_unit/models/crm_lead.py
AaronHForgeFlow: It's working with the domain, thank you @elicoidal. And thanks also to @SerpentCS for the help

JordiBForgeFlow:
:+1:

JordiBForgeFlow (approved):

lfreeke:
:+1: (test)

PR #27, opened 2016-09-15
[ADD][9.0] crm claim operating unit

AaronHForgeFlow:
- Adds the Operating Unit (OU) to the CRM Claim.
- User can only view and manage the claims associated to the warehouse of the Operating Units that he has access to.


JordiBForgeFlow (approved):
:+1:

serpentcs-dev1:
Tested and works fine :+1:

serpentcs-dev1 (approved):

crm_claim_operating_unit/tests/test_crm_claim_operating_unit.py
```diff
@@ -54,7 +52,7 @@ def _create_user(self, login, groups, company, operating_units,
 
     def _create_crm_claim(self, uid, operating_unit):
         """Creates a CRM Claim."""
-        claim = self.crm_claim_model.sudo(uid).create({
+        claim = self.crm_claim_model.create({
```
LoisRForgeFlow: Why do you remove the sudo?

crm_claim_operating_unit/models/crm_claim.py
```diff
@@ -14,9 +13,10 @@ class CRMClaim(models.Model):
 
     @api.model
     def _default_operating_unit(self):
-        if self._defaults['team_id'](self):
-            return self.env['crm.team'].browse(self._defaults['team_id'](
-                self)).operating_unit_id
+        team_id = self.env['crm.team']._get_default_team_id()
+        team = self.env['crm.team'].sudo().browse(team_id)
```
LoisRForgeFlow: Why this other sudo?

gmeficent (approved):
Functional and code review :+1:

Only **one thing to mention**, to make sure it works as intended.

I created a team and a claim with the same OU and then I deleted that OU from my access right.
After that I still can see and edit the claims that I created with that OU and I still belong to that OU team.

MiquelRForgeFlow (approved):
It's squashed. Ready to be merged!!

LoisRForgeFlow (approved):

aheficent:
Not ready to merge. Comments from @gmeficent  unnatended

AaronHForgeFlow:
@gmeficent I think that behavior is correct. If the operating unit is deleted then the sales team has no operating unit, so everybody can access. For example, when deleting an operating unit all the sales for that operating unit are not deleted, the operating unit is set to null in this case.

PR #30, opened 2016-09-15
[ADD][9.0] hr expense operating unit

AaronHForgeFlow:
- Adds the Operating Unit (OU) to the Expense.
- Security rules are defined to ensure that users can only see the Expense of that Operating Units in which they are allowed access to.
- Adds Operating Unit (OU) to the account moves while generating accounting entries from the expense.


hr_expense_operating_unit/models/hr_expense.py
```diff
@@ -0,0 +1,24 @@
...
+
+class HrExpenseExpense(models.Model):
+
+    _inherit = 'hr.expense'
+
+    operating_unit_id = fields.Many2one('operating.unit', 'Operating Unit',
+                                        default=lambda self:
+                                        self.env['res.users'].
+                                        operating_unit_default_get(self._uid))
+
```
serpentcs-dev1, line 18: Its better if we add the constraint for company in OU and Expense :

```
@api.multi
@api.constrains('operating_unit_id', 'company_id')
def _check_company_operating_unit(self):
        for rec in self:
            if rec.company_id and rec.operating_unit_id and \
                    rec.company_id != rec.operating_unit_id.company_id:
                raise Warning(_('The Company in the Expense and in '
                                'the Operating Unit must be the same.'))

Comment (AaronHForgeFlow) [reply hr_expense_operating_unit/models/hr_expense.py:18]: Done, Thank you @darshan-serpent

**Comment by JordiBForgeFlow** [comment]
@lreficent @mreficent can you review?

File: hr_expense_operating_unit/security/hr_expense_security.xml [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+    <record id="ir_rule_hr_expense_allowed_operating_units"
+            model="ir.rule">
+        <field name="model_id" ref="hr_expense.model_hr_expense"/>
+        <field name="domain_force">['|',('operating_unit_id','=',False),('operating_unit_id','in',[g.id for g in user.operating_unit_ids])]</field>
+        <field name="name">Expenses from allowed operating units</field>
+        <field name="global" eval="True"/>
+        <field eval="0" name="perm_unlink"/>
+        <field eval="0" name="perm_write"/>
+        <field eval="1" name="perm_read"/>
+        <field eval="0" name="perm_create"/>
```
sbidoul: IMO it must be 1 for unlink, write and create too, otherwise the record rule won't apply for these operation, so a user could changes records by guessing their ids.

LoisRForgeFlow: I agree with @sbidoul

hr_expense_operating_unit/tests/test_hr_expense_operating_unit.py
```diff
@@ -0,0 +1,113 @@
...
+    def _post_journal_entries(self, expense):
+        """Approves the Expense and creates accounting entries."""
+        expense.submit_expenses()
+        expense.approve_expenses()
+        expense.action_move_create()
+
+    def test_security(self):
+        # User 2 is only assigned to Operating Unit B2C, and cannot
+        # Access Expenses of Main Operating Unit.
+        record = self.hr_expense_model.sudo(self.user2.id).search(
```
LoisRForgeFlow, line 102: It would be good to expand the test with creation or unlink. Following the fix proposed by @sbidoul

LoisRForgeFlow: Done.

lreficent:
I didn't expand the rule to create operation because I think you cannot do a harmful with this operation and also maybe some manager wants to create records for another op.

LoisRForgeFlow:
I didn't expand the rule to create operation because I think you cannot do a harmful action with this operation and also maybe some manager wants to create records for another op.

LoisRForgeFlow (approved):
from my side :+1:

AaronHForgeFlow:
Not working anymore on this

PR #31, opened 2016-09-15
[9.0][stock_operating_unit]

JordiBForgeFlow:
Supersedes https://github.com/OCA/operating-unit/pull/20


JordiBForgeFlow:
cc @aheficent Can you review?

AaronHForgeFlow:
IMHO this is ready to be merged :+1:

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,146 @@
...
+
+    operating_unit_id = fields.Many2one('operating.unit',
+                                        'Operating Unit')
+
+    @api.multi
+    @api.constrains('operating_unit_id')
+    def _check_warehouse_operating_unit(self):
+        for rec in self:
+            warehouse_obj = self.env['stock.warehouse']
+            warehouse = warehouse_obj.search([('wh_input_stock_loc_id', 'in',
```
angelmoya: you could use:
 '=', rec.id
instead of:
'in', rec.ids
rec is one, so you do not need to use "in"

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,146 @@
...
+                                               rec.ids)])
+            for w in warehouse:
```
angelmoya: you could use one for w in warehouse and test tree conditions instead create tree for for each condition.

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,146 @@
...
+                Stock Picking and in the Operating Unit must be the same.'))
+
+    @api.multi
+    @api.constrains('operating_unit_id', 'move_lines')
+    def _check_stock_move_operating_unit(self):
+        for rec in self:
+            if not rec.operating_unit_id:
+                return True
+            operating_unit = rec.operating_unit_id
+            for sm in rec.move_lines:
```
angelmoya: It would be nice to use more clean names.
sm -> move or stock_move

angelmoya (commented):
Code review done, I will test on runbot now.

Some remarks but it's ok

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,146 @@
...
+    def _check_company_operating_unit(self):
+        for rec in self:
+            if rec.company_id and rec.operating_unit_id and \
+                    rec.company_id != rec.operating_unit_id.company_id:
+                raise UserError(_('Configuration error!\nThe Company in the\
+                Stock Picking and in the Operating Unit must be the same.'))
+
+    @api.multi
+    @api.constrains('operating_unit_id', 'move_lines')
+    def _check_stock_move_operating_unit(self):
```
angelmoya: Why do not put this constraint on stock moves?

angelmoya:
On stock picking lines, I think that:
- Source Location Operating Unit 
- Dest. Location Operating Unit 

are to long labels for user interface

stock_operating_unit/model/stock.py
AaronHForgeFlow: Yes, but as rec.ids is a list i got to use rec.ids[0]

stock_operating_unit/model/stock.py
AaronHForgeFlow: All right. Thank you Ángel

stock_operating_unit/model/stock.py
AaronHForgeFlow: Done

AaronHForgeFlow:
About the long names I think that there are long, however if we shorten them it will be confusing.

manuelfcordoba:
Creating a delivery order I noticed that the Requesting Operating Unit by default is set to the Main Operating Unit. I think it should be =false by default.

Creating a delivery order throw the kan ban view of an specific Picking type and changing the Picking type in the form view generated an inconsintence on the Picking type of the moves.

AaronHForgeFlow:
Travis Lint is failing because the onchange_picking_type method is in v7 in the odoo core, so it has to be overwritten this way.

LoisRForgeFlow (commented):
When accessing to the stock locations as stock manager with Demo User and the B2B Operating Unit assigned, the next access error pops:
image
I think the stock manager of a OU should have access to the locations of his/her warehouse at least.

AaronHForgeFlow:
I also realize that when creating stock pickings, if the user changes the destination location the system does not change the requesting operating unit and the stock picking

AaronHForgeFlow:
Regarding my last comment. The Picking type is the one that controls the operating unit of the picking. If the user changes the destination location the system automatically changes the destination location operating unit, so it was already correct.

AaronHForgeFlow (approved):
LGTM now :+1:

LoisRForgeFlow (approved):
Tested on runbot + code review :+1:

AaronHForgeFlow:
@angelmoya can you review?

stock_operating_unit/__openerp__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# © 2016 Eficent Business and IT Consulting Services S.L.
+# © 2016 Serpent Consulting Services Pvt. Ltd.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Stock with Operating Units",
+    "summary": "An operating unit (OU) is an organizational entity part of a\
+        company",
```
guewen: The following form would be better:

```
    "summary": "An operating unit (OU) is an organizational entity part of a "
               "company",
```

Because ithe current form includes a lot of spaces before "company".

stock_operating_unit/demo/stock_demo.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="1">
```
guewen: `<openerp><data noupdate="1">` can be replaced by `<odoo noupdate="1">`

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,169 @@
+# -*- coding: utf-8 -*-
+# © 2016 Eficent Business and IT Consulting Services S.L.
+# © 2016 Serpent Consulting Services Pvt. Ltd.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from openerp.tools.translate import _
+from openerp import api, fields, models
```
guewen: `_` can be imported from the same package: `from openerp import _, api, fields, models`

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,169 @@
...
+from openerp.exceptions import Warning as UserError
```
guewen: Better: `from openerp.exceptions import UserError`. `Warning` is deprecated.

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class StockWarehouse(models.Model):
+    _inherit = "stock.warehouse"
+
+    operating_unit_id = fields.Many2one('operating.unit',
+                                        'Operating Unit',
+                                        default=lambda self:
+                                        self.env['res.users'].
+                                        operating_unit_default_get(self._uid))
```
guewen: `self.env.uid` is a cleaner (no access to a "private" attribute).

You would have less line breaks with such indentation:

```
        comodel_name='operating.unit',
        string='Operating Unit',
        default=lambda self: self.env['res.users'].
                operating_unit_default_get(self.env.uid))
    )
```

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @api.multi
+    @api.constrains('operating_unit_id', 'company_id')
+    def _check_company_operating_unit(self):
+        for rec in self:
+            if rec.company_id and rec.operating_unit_id and rec.company_id \
+                    != rec.operating_unit_id.company_id:
```
guewen: `\` should be avoided when it parenthesis can be used:

```python

if (rec.company_id and rec.operating_unit_id and 
    rec.company_id != rec.operating_unit_id.company_id):



```

stock_operating_unit/model/stock.py
```diff
@@ -0,0 +1,169 @@
...
+                raise UserError(_('Configuration error!\nThe Company in the\
+                Stock Warehouse and in the Operating Unit must be the same.'))
```
guewen: This string contains a lot of spaces which is then displayed weirdly in the error message. You should use the string concatenation:

```python

raise UserError(
    _('Configuration error\nThe Company in the '
      'Stock Warehouse and in the Operating Unit must be the same.')
)

```

(note: I also removed the exclamation point in the message which is not recommended (https://cybertext.wordpress.com/2011/11/18/error-messages-that-make-you-feel-stupid/)

guewen (commented):
I started to leave comments but I'll submit a PR to make the changes directly.

AaronHForgeFlow:
@guewen should I close this?

guewen:
@aheficent nope, here is my proposal that you can merge in yours if you agree https://github.com/Eficent/operating-unit/pull/64

AaronHForgeFlow:
All right, Thank you @guewen

guewen (approved):
Thanks!
There are enough approved reviews to merge.

Note: I have the rights to merge but I don't want to walk over the Operating Units PSC

guewen:
Merging, it will unblock `procurement_operating_unit` (which in turns will unblock `purchase_operating_unit` once merged).

PR #35, opened 2016-10-05
[9.0] sales_team_operating_unit

JordiBForgeFlow:


crm_operating_unit/models/crm_lead.py
```diff
@@ -3,14 +3,36 @@
...
+                                        default=_get_default_operating_unit)
+
+    @api.onchange('team_id')
+    def onchange_team_id(self):
+        self.operating_unit_id = self.team_id.operating_unit_id
+
+    @api.multi
+    @api.constrains('team_id', 'operating_unit_id')
+    def _check_team_operating_unit(self):
+        for rec in self:
```
pedrobaeza: This constraint has no sense. Makes a related field if you want this behavior.

sales_team_operating_unit/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+=========================
+Sales Team Operating Unit
+=========================
+
+This module introduces the following features:
+
+* Adds the Operating Unit (OU) to the Sales Team.
+
+* The user’s default Operating Unit (OU) is proposed at the time of creating
```
pedrobaeza: s/`/'

crm_operating_unit/models/crm_lead.py
```diff
@@ -3,14 +3,36 @@
...
-from openerp import fields, models
+from openerp import api, fields, models, _
+from openerp.exceptions import UserError
 
 
 class CRMLead(models.Model):
 
     _inherit = 'crm.lead'
 
+    @api.model
```
pedrobaeza, line 12: You can't add this code here, as there's no dependency to the other module. Do it in the proper module.

sales_team_operating_unit/models/crm_team.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# © 2015 Eficent Business and IT Consulting Services S.L. -
+# Jordi Ballester Alomar
```
pedrobaeza: Don't include 2 copyright lines. Make it in one

sales_team_operating_unit/tests/test_crm_team_operating_unit.py
```diff
@@ -0,0 +1,67 @@
...
+class TestSaleTeamOperatingUnit(common.TransactionCase):
+
+    def setUp(self):
+        super(TestSaleTeamOperatingUnit, self).setUp()
+        self.res_users_model = self.env['res.users']
+        self.crm_team_model = self.env['crm.team']
+        # Groups
+        self.grp_sale_mngr = self.env.ref('base.group_sale_manager')
+        self.grp_user = self.env.ref('base.group_user')
+        # Company
```
pedrobaeza, line 17: There's no need for these comments. You can use significant variable names instead.

pedrobaeza (changes requested):

crm_operating_unit/models/crm_lead.py
JordiBForgeFlow: The team_id is not a required field in the lead. You might have a lead for an operating unit not associated to a sales team.

JordiBForgeFlow: The team_id is not a required field in the lead. You might have a lead for an operating unit not associated to a sales team.

pedrobaeza: Yeah, but if present, you're not allowing to change it, so it doesn't make too much sense to have it. Also the message is not correct. It says 'company' instead of operating unit.

AaronHForgeFlow:
As long as crm depends on sales_team, crm_operating_unit depends on sales_team_operating_unit. Consequently this PR changes crm_operating_unit to adapt the tests.

alan196 (approved):

crm_operating_unit/models/crm_lead.py
AaronHForgeFlow: Now, there's dependency. I think it's ok, isn't it?

AaronHForgeFlow:
I Don't know why coverage decreases, IMO tests cover all what this module introduces.

pedrobaeza:
Default method is not being tested:

seleccion_009

AaronHForgeFlow:
Thank you @pedrobaeza

guewen:
@pedrobaeza can you reconsider your review? thanks

guewen (approved):

JordiBForgeFlow:
The review from @pedrobaeza refers to the module crm, that has already been merged.

PR #40, opened 2016-11-16
[FIX][account_operating_unit]create method of account move expect new…

IsabelEsparza:
… parameter when cash basis is installed

Add parameter apply_taxes on method create, for the validation of any account move, because when you install cash basis module it needs that parameter


guewen:
Hello @IsabelEsparza 

It seems you have included a commit (ad43b29) that shouldn't be there, can you check it?
Also, did you signed the CLA since you've created your pull request?

Thanks

account_operating_unit/models/account_move.py
```diff
@@ -16,12 +16,12 @@ class AccountMoveLine(models.Model):
...
 
     @api.model
-    def create(self, vals):
+    def create(self, vals, apply_taxes=False):
         if vals.get('move_id', False):
             move = self.env['account.move'].browse(vals['move_id'])
             if move.operating_unit_id:
                 vals['operating_unit_id'] = move.operating_unit_id.id
-        return super(AccountMoveLine, self).create(vals)
+        return super(AccountMoveLine, self).create(vals, apply_taxes)
```
guewen, line 24: named argument should be propagated using the keyword:

```python
return super(AccountMoveLine, self).create(vals, apply_taxes=apply_taxes)
```

guewen:
I hope you don't mind, I corrected your PR in #43 so I close this one.

PR #45, opened 2017-01-26
[MIG] operating_unit to v10.0

MiquelRForgeFlow:
Operating Unit
==============

An operating unit (OU) is an organizational entity part of a company, with
separate management ownership. Management by OU is aimed to introduce the
following features:

* Data partitioning from other OU.
* Own sequencing schemes.
* User access administration for data processing and reporting.
* Product and customer independence of OU.
* OU specific P&L and Balance sheet.

PR #49, opened 2017-01-26
10.0 mig procurement operating unit

AaronHForgeFlow:
Procurement Operating Unit
==========================

- Implements global security rules on procurement orders so that a user can
only read procurement orders where the location is linked to an operating unit
that the user has access to.

Installation
============

No external library is used.

Configuration
=============

Usage
=====

This module defines the operating unit entity and the user's security rules.
Other modules extend the standard Odoo apps with the OU.

# Depends on https://github.com/OCA/operating-unit/pull/51


LoisRForgeFlow:
depends on https://github.com/OCA/operating-unit/pull/45 and https://github.com/OCA/operating-unit/pull/51

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

PR #50, opened 2017-01-26
[MIG] account_operating_unit to v10.0

LoisRForgeFlow:
Accounting with Operating Units
===============================

This module allows a company to manage the accounting based on Operating
Units (OU's).

* The financial reports (Trial Balance, P&L, Balance Sheet), allow to report
  the balances of one or more OU's.

* If a company wishes to report Balance Sheet and P&L accounts based on
  OU's, they should indicate at company level that the OU's are
  self-balanced, and the corresponding Inter-Operating Unit clearing account.
  The Chart of Accounts will always be balanced, for each Operating Unit.

* A company considering Operating Unit as applicable to report only profits
  and losses will not need to set the OU's as self-balanced.

* The self-balancing of Operating Unit is ensured at the time of posting a
  journal entry. In case that the journal involves posting of items in
  separate Operating Units, new journal items will be created, using the
  Inter-Operating Unit clearing account, to ensure that each OU is going to
  be self-balanced for that journal entry.

* Adds the Operating Unit (OU) to the invoice. A user can choose what OU to
  create the invoice for.

* Adds the Operating Unit (OU) to payments and payment methods. The operating
  unit of a payment will be that of the payment method choosen.

* Implements security rules at OU level to invoices, payments and journal
  items.

### DIFF
--- a/account_operating_unit/models/account_journal.py
+++ b/account_operating_unit/models/account_journal.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2016 Eficent Business and IT Consulting Services S.L.
+# © 2016-17 Eficent Business and IT Consulting Services S.L.
 # © 2016 Serpent Consulting Services Pvt. Ltd.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 from openerp import api, fields, models, _
--- a/account_operating_unit/README.rst
+++ b/account_operating_unit/README.rst
@@ -30,7 +30,7 @@ Units (OU's).
   create the invoice for.
 
 * Adds the Operating Unit (OU) to payments and payment methods. The operating
-  unit of a payment will be that of the payment method choosen.
+  unit of a payment will be that of the payment method chosen.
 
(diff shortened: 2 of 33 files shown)


lreficent:
depends on https://github.com/OCA/operating-unit/pull/45

LoisRForgeFlow:
depends on https://github.com/OCA/operating-unit/pull/45

JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:

account_operating_unit/models/account_journal.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2016 Eficent Business and IT Consulting Services S.L.
+# © 2016-17 Eficent Business and IT Consulting Services S.L.
 # © 2016 Serpent Consulting Services Pvt. Ltd.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 from openerp import api, fields, models, _
```
serpentcs-dev1, line 5: should have used odoo instead

PR #70, opened 2017-05-24
[FIX][10.0]Singleton error when paying more than one invoice

AaronHForgeFlow:
I suggest to consider the OU of the invoice when there's just one invoice. Otherwise the OU of the journal is considered.


AaronHForgeFlow:
cc @eficent

sbidoul (approved):

LoisRForgeFlow (approved):
:+1:

MiquelRForgeFlow (approved):
:1st_place_medal: :+1:

AaronHForgeFlow:
@jbeficent @guewen  Your review would be appreciated

PR #76, opened 2017-06-19
[9.0][FIX]operating unit when onchange picking type

AaronHForgeFlow:
The onchange method is not compatible with readonly fields. 

When changing the picking type the operating unit of the warehouse was not been stored.

As there's  a constraint that checks for incompatibilities between the operating unit in the warehouse and in the picking I think the field operating_unit_id in the picking can be editable.

This PR Includes another small fix in the method  `_check_warehouse_operating_unit`

### DIFF
--- a/stock_operating_unit/model/stock.py
+++ b/stock_operating_unit/model/stock.py
@@ -48,14 +48,6 @@ class StockLocation(models.Model):
                     raise UserError(_('Configuration error\nThis location is '
                                       'assigned to a warehouse that belongs to'
                                       ' a different operating unit.'))
-                if self.operating_unit_id != w.operating_unit_id:
-                    raise UserError(_('Configuration error\nThis location is '
-                                      'assigned to a warehouse that belongs to'
-                                      ' a different operating unit.'))
-                if rec.operating_unit_id != w.operating_unit_id:
-                    raise UserError(_('Configuration error\nThis location is'
-                                      ' assigned to a warehouse that belongs'
-                                      ' to a different operating unit.'))
 
     @api.multi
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
@eficent @darshan-serpent

serpentcs-dev1 (approved):
Code Review and tested on runbot 

LGTM :+1: 

Yes it seems that when a field is set as readonly, the values will not be stored in the database.

AaronHForgeFlow:
Yes @darshan-serpent when the field is readonly it won't store the field even if the readonly is just in the view :open_mouth:

MiquelRForgeFlow (approved):

stock_operating_unit/model/stock.py
```diff
@@ -48,7 +48,7 @@ def _check_warehouse_operating_unit(self):
                     raise UserError(_('Configuration error\nThis location is '
                                       'assigned to a warehouse that belongs to'
                                       ' a different operating unit.'))
-                if self.operating_unit_id != w.operating_unit_id:
+                if rec.operating_unit_id != w.operating_unit_id:
```
MiquelRForgeFlow: I think that `if` is duplicated.

MiquelRForgeFlow (changes requested):

AaronHForgeFlow:
Thank you for your review @mreficent Your request have been attended :call_me_hand:

MiquelRForgeFlow (approved):

AaronHForgeFlow:
I just squashed the commits. No code affected.

LoisRForgeFlow (approved):
:+1:

PR #82, opened 2017-07-11
[10.0] [ADD] hr_payroll_account_operating_unit

serpentcs-dev1:
HR Payroll Account with Operating Units
=======================================

* Adds Operating Unit (OU) to the account moves and its lines created by the payslip, based on the Operating Unit (OU) defined in the Employee's Contract.


nikul-serpentcs:
@aheficent Could you please review code.

hr_payroll_account_operating_unit/models/hr_payslip.py
```diff
@@ -0,0 +1,42 @@
...
+                                   slip.contract_id.operating_unit_id.id})
+        return res
+
+    @api.multi
+    def action_payslip_done(self):
+        OU = None
+        for slip in self:
+            # Check that all slips are related to contracts
+            # that belong to the same OU.
+            if OU:
```
AaronHForgeFlow: Why don't compare directly?

hr_payroll_account_operating_unit/models/hr_payslip.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class HrPayslip(models.Model):
+
+    _inherit = 'hr.payslip'
+
+    @api.multi
+    def write(self, vals):
+        res = super(HrPayslip, self).write(vals)
+        if 'move_id' in vals and vals['move_id']:
```
AaronHForgeFlow: you can use `vals.get('move_id', False)` but it's the same

hr_payroll_account_operating_unit/models/hr_payslip.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @api.multi
+    def action_payslip_done(self):
+        OU = None
+        for slip in self:
+            # Check that all slips are related to contracts
+            # that belong to the same OU.
+            if OU:
+                if slip.contract_id.operating_unit_id.id != OU:
+                    raise UserError(_('Configuration error!\nThe Contracts must\
```
AaronHForgeFlow: The \n is not needed just close the quotes and continue in the next line

hr_payroll_account_operating_unit/models/hr_payslip.py
```diff
@@ -0,0 +1,42 @@
...
+    def write(self, vals):
+        res = super(HrPayslip, self).write(vals)
+        if 'move_id' in vals and vals['move_id']:
+            for slip in self:
+                if slip.contract_id and slip.contract_id.operating_unit_id:
+                    slip.move_id.write({'operating_unit_id':
+                                        slip.contract_id.operating_unit_id.id})
+                    if slip.move_id.line_ids:
+                        slip.move_id.line_ids.\
+                            write({'operating_unit_id':
```
AaronHForgeFlow, line 27: Won't be better to include this in the account_move class?

AaronHForgeFlow (commented):
Only code review :neutral_face:

hr_payroll_account_operating_unit/models/hr_payslip.py
nikul-serpentcs: I didn't get you?! Do you want me to move this whole thing to while account_move are created?

hr_payroll_account_operating_unit/models/hr_payslip.py
nikul-serpentcs: This is comparing the OU in contracts between the Payslip. I don't see the use case for this though.

Btw, what changes do you want me to make?

hr_payroll_account_operating_unit/models/hr_payslip.py
AaronHForgeFlow: In v9 it was done by passing the operating unit in the context and then overriding the create method in the account_move. But maybe this is good so leave it like this.

nikul-serpentcs: Ok, Thanks

hr_payroll_account_operating_unit/models/hr_payslip.py
AaronHForgeFlow: I was wondering if there is a way to map the contracts by different OU and if there was more than one raise the error. But I'm not really sure and this works. It's ok to me.

AaronHForgeFlow:
Code looks now. We have to wait to the dependencies are merged.

AaronHForgeFlow:
Superseded by #127

mreficent:
Can be closed. Superseeded by https://github.com/OCA/operating-unit/pull/127.

PR #89, opened 2017-09-26
[10.0] [ADD]  sale_crm_operating_unit

nikul-serpentcs:
Sale CRM Operating Unit
=======================

* Passes the Operating Unit (OU) from Opportunity to Quotation.


nikul-serpentcs:
@aheficent Could you please review.

nikul-serpentcs:
@aheficent Could you please review.

sale_crm_operating_unit/models/sale.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.constrains('opportunity_id', 'operating_unit_id')
+    def _check_sale_operating_unit(self):
+        if self.opportunity_id and self.opportunity_id.operating_unit_id and\
+                self.opportunity_id.\
+                operating_unit_id != self.operating_unit_id:
+            raise ValidationError(_('''Configuration error! The Operating Unit
```
AaronHForgeFlow: Don't need to yet at users, because that may hurt some feelings :cry:

AaronHForgeFlow (changes requested):
Tested on runbot. The operating unit is passed from the lead to the opportunity but not form the opportunity to the quotation.

nikul-serpentcs:
@aheficent Could you please review.

AaronHForgeFlow:
Not sure is runbot was rebuilt or not but still the operating unit is not passed from the opportunity to the quotation

nikul-serpentcs:
@aheficent Could you please check functionality, runbot is fine.

AaronHForgeFlow (approved):
:+1: It works ok.

gdgellatly:
@nikul-serpentcs @aheficent i was asked to do some reviews which I'm of course happy to do although the code style in this repo is a little different to what I'm used to. 

I have one big complaint with this PR.  It unnecessarily overrides the context of the sale actions.  Before we had web_context_tunnel but now you get one shot and I think it should be avoided wherever possible as it widens the universe of incompatible modules.  And unfortunately these actions are widely overridden as a way to pass information from opportunity to quotation.

I think a better way is to either use an onchange opportunity_id (slight change of behaviour) or else detect the existing opportunity in context in the call (or the return from super) in fields_default_get (identical behaviour) and set it that way. (I see its not in your context, I can't remember, maybe it is in the action context, but it is passed).

AaronHForgeFlow:
@gdgellatly Is there a way to extend the context? If it's not possible I'd use fields_default_get method to achieve the same functionality. @nikul-serpentcs

gdgellatly:
@aheficent unfortunately not quite.  Nearly, very nearly. Attribute overrides support an add and remove argument.  It works for some attributes, but not domains, attrs or contexts (anything that gets eval'ed basically).

See https://github.com/odoo/odoo/issues/17960

AaronHForgeFlow:
In that case this module should not overwrite the context this way.

pedrobaeza:
Use module https://github.com/OCA/server-tools/tree/10.0/base_view_inheritance_extension that allows to add values to the context.

gdgellatly:
@pedrobaeza Oh my thats beautiful, whats it doing hidden away in server tools.  That module nearly needs to be in the contributing guideline for extending views.

pedrobaeza:
If you consider it, make a PR in the guidelines :wink:

AaronHForgeFlow:
I agree then to make this module depends on `base_view_inheritance_extension`  and adapt this module to work with that. Can you do the thing @nikul-serpentcs?

AaronHForgeFlow:
Continued here #89

PR #90, opened 2017-09-28
[FIX][10.0]stock_operating_unit

AaronHForgeFlow:
**Requesting operating unit in manual picking never been set**

Current behavior: When selecting the picking type in the picking the requesting operating unit is automatically set. When saving, the picking the requesting operating unit is missing.

Expected behavior: The requesting operating unit remains there.

Solution: Set the requesting operating unit as editable. Currently, there's a constraint that prevents for inconsistencies so there's no risk of inconsistency data.

### DIFF
--- a/stock_operating_unit/__manifest__.py
+++ b/stock_operating_unit/__manifest__.py
@@ -7,7 +7,7 @@
     "name": "Stock with Operating Units",
     "summary": "An operating unit (OU) is an organizational entity part of a "
                "company",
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
@eficent

LoisRForgeFlow (approved):
Tested on runbot, LTGM

MiquelRForgeFlow (approved):

PR #97, opened 2017-10-08
[ADD] [10.0] hr_employee_operating_unit

Matiar:

1. Adds the Operating Unit (OU) to the Employee.
2. The user’s default Operating Unit (OU) is proposed at the time of creating the Employee.
3. Security rules are defined to ensure that users can only see the Employee of that Operating Units in which they are allowed access to.


Matiar:
@aheficent Could you please review ?

AaronHForgeFlow (changes requested):
Hi @Matiar Thanks for your module. A couple of things:

- As there's a field user_id in the employee, if the employee has an user associated, then the employee should have an operating unit from the operating unit of the users.
- IMHO an employee may belong to different operating units so the field may be One2many. The problem here is that it may be inconsistencies between the operating unit of the employee and the operating unit of the related user.
- Please attend travis and also fix the description in this PR

Matiar:
Hello @Aarón Henríquez,

Thanks for your valuable feedback. I'll fix as you suggested and get back to you.

Matiar:
[[FIX] When an employee tag with an user account, the default OU is set with the default OU of user account](https://github.com/OCA/operating-unit/pull/97/commits/85c0b1dec60c4ba4cbe69736ae3eb166c3b2efcb)

Matiar:
Hello Aarón Henríquez,

Thanks for your valuable feedback.

As you suggested, when an employee will associate with an user account the operating unit of employee is set with the operating unit of the associated user.

But also you suggest an one2many relationship between employee and operating unit, I didn't yet implement that. I have a question on that as you know the employee has a relationship between department and that is Many2One; Then will it not confusing to maintain One2Many relationship.

Would you please review and let me know your valuable feedback ?

Matiar:
@aheficent Could you please review ?

AaronHForgeFlow:
Hi @Matiar the question is, can an employee belong to different operating units? If the answer it's true then the field have to be Many2many.

AaronHForgeFlow:
I think it's not necessary to have a default_operating_unit field in the employee. I'd leave just the many2many relationship. Don't you agree?

hr_employee_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,47 @@
...
+                                 default=lambda self: self.env.user.company_id)
+
+    operating_unit_ids = fields.Many2many('operating.unit',
+                                          'operating_unit_emp_rel',
+                                          'emp_id', 'ou_id', 'Operating Units',
+                                          default=_get_operating_units)
+
+    @api.onchange('user_id')
+    def _onchange_user(self):
+        super(HREmployee, self)._onchange_user()
```
AaronHForgeFlow: Hi @Matiar  Thank you for doing the thing but I was wrong. An employee may belong to different operating units but as a user manage different operating units. So probably this is not needed now. I'm sorry for the misunderstanding.

hr_employee_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,47 @@
...
+    @api.model
+    def operating_unit_default_get(self, uid2):
+        if not uid2:
+            uid2 = self._uid
+        user = self.env['res.users'].browse(uid2)
+        return user.default_operating_unit_id
+
+    @api.model
+    def _get_operating_unit(self):
+        return self.operating_unit_default_get(self._uid)
```
AaronHForgeFlow: Is it possible to achieve the same result in a single method?

Matiar:
Hello @aheficent ,

Thanks for your overall feedback and reviewing the PR. I'm not sure how to fix codecov/project & codecov/patch failure issue. Could you please share some source or link to understand the problem and fixing it.

AaronHForgeFlow:
Hi @Matiar,

You need to add some tests in order to turn green codecov. For example you can test that users cannot see the employees the are not allowed to. You can also test the default method.Are you familiarized with unit test?

Matiar:
Hello @aheficent, 

Thanks for your valuable feedback and keeping your patience to accept my mistakes. I added the unit tests as you suggested. Could you please review again ?

hr_employee_operating_unit/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2017 Genweb2 Limited - Matiar Rahman
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from . import models
+from . import tests
```
AaronHForgeFlow: Importing tests is not needed

AaronHForgeFlow (commented):
Code looks nice. Only a minor thing

hr_employee_operating_unit/__init__.py
Matiar: Hi @aheficent ,

I fixed that. Thanks.

AaronHForgeFlow (commented):
It works fine. Just a little thing. Can you remove the Create and edit from the widget?
image

BTW, have you signed the CLA?

Matiar:
Hello @aheficent ,

Thanks for your review and valuable feedback, But IMHO, the change you suggested belongs to the feature of HR module, do you think this module (hr_employee_operating_unit) is the right place to 
implement your suggested change ?

JordiBForgeFlow:
He meant from the operating unit field.

Matiar:
Hello @aheficent , 

Thanks for your feedback. I incorporated your suggested change. Could you please review again ? 
And yes, I signed the CLA couple of months back.

hr_employee_operating_unit/security/security.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
+<data noupdate="0">
```
AaronHForgeFlow: data it's not necessary anymore. Use <odoo noupdate="0">

hr_employee_operating_unit/views/hr_views.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<openerp>
```
AaronHForgeFlow: odoo

hr_employee_operating_unit/views/hr_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                        <field name="operating_unit_ids"
+                               domain="[('company_id', '=', company_id)]"
+                               widget="many2many_tags"
+                               options="{'no_create': True}" />
+                    </xpath>
+                </data>
+            </field>
+        </record>
+
+        <!-- Employee -->
```
AaronHForgeFlow: Not needed IMO

AaronHForgeFlow (commented):
Tested on runbot and working fine. Just some minor comments to the code.

Matiar:
Hello @aheficent , 

Thanks again for your time and feedback. I've fixed your suggested changes. Take a look at your convenient  time.

hr_employee_operating_unit/security/security.xml
AaronHForgeFlow: Sorry. I meant `<odoo noupdate="0">` but it does not matter because that it's the same as `<odoo>`

AaronHForgeFlow (approved):
:+1: Functional + code. Thanks @Matiar for attending all the comments.

Matiar:
Hello @aheficent ,

Thanks for your time and approve the PR. Could you please inform me what will be my next step to merge the PR with OCA branch or when it will be merged ?

AaronHForgeFlow:
Hi @Matiar we need at least two more approvals. Normally it takes some time but eventually it'll be merged :smile:

hr_employee_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import api, fields, models
+
+
+class HREmployee(models.Model):
+
+    _inherit = 'hr.employee'
+
+    @api.model
+    def _get_operating_unit(self):
+        user = self.env['res.users'].browse(self._uid)
```
grindtildeath, line 13: No need to browse here, you can do `return self.env.user.default_operating_unit_id`

hr_employee_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,26 @@
...
+        return user.default_operating_unit_id
+
+    @api.model
+    def _get_operating_units(self):
```
grindtildeath, line 17: I suppose you can get rid of this method and call directly `_get_operating_unit` ?

hr_employee_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,26 @@
...
+        return self._get_operating_unit()
+
+    company_id = fields.Many2one('res.company', string='Company',
+                                 default=lambda self: self.env.user.company_id)
+
+    operating_unit_ids = fields.Many2many('operating.unit',
+                                          'operating_unit_emp_rel',
+                                          'emp_id', 'ou_id', 'Operating Units',
+                                          default=_get_operating_units)
```
grindtildeath, line 26: Please use lambda to proxy the call to `_get_operating_unit`

hr_employee_operating_unit/tests/test_employee_operating_unit.py
```diff
@@ -0,0 +1,69 @@
...
+                                       self.company, [self.b2c_OU])
+
+        # Create Employee
+        self.emp1 = self._create_employee(self.user1.id)
+        self.emp2 = self._create_employee(self.user2.id)
+
+    def _create_user(self, login, groups, company, operating_units):
+        """ Create a user. """
+        group_ids = [group.id for group in groups]
+        user = self.res_users_model.create({
```
grindtildeath, line 38: Here would be nice to add `with_context('no_reset_password': True, 'mail_create_nosubscribe': True)` to avoid sending unnecessary emails during the test.

hr_employee_operating_unit/tests/test_employee_operating_unit.py
```diff
@@ -0,0 +1,69 @@
...
+                                                  self.grp_user],
+                                       self.company, [self.b2c_OU])
+
+        # Create Employee
+        self.emp1 = self._create_employee(self.user1.id)
+        self.emp2 = self._create_employee(self.user2.id)
+
+    def _create_user(self, login, groups, company, operating_units):
+        """ Create a user. """
+        group_ids = [group.id for group in groups]
```
grindtildeath, line 37: You can remove this line and do `groups.ids` to get the same list.

hr_employee_operating_unit/tests/test_employee_operating_unit.py
```diff
@@ -0,0 +1,69 @@
...
+        user = self.res_users_model.create({
+            'name': login,
+            'login': login,
+            'password': 'demo',
+            'email': 'test@yourcompany.com',
+            'company_id': company.id,
+            'company_ids': [(4, company.id)],
+            'operating_unit_ids': [(4, ou.id) for ou in operating_units],
```
grindtildeath, line 45: Shouldn't `[6, 0, operating_units.ids]` do the same ?

hr_employee_operating_unit/tests/test_employee_operating_unit.py
```diff
@@ -0,0 +1,69 @@
...
+        })
+        return employee
+
+    def test_employee(self):
+        # User 2 is only assigned to B2C Operating Unit, and cannot
+        # access Employee for Main Operating Unit.
+
+        emp = self.hr_employee_model.sudo(self.user2.id).search(
+            [('id', '=', self.emp1.id),
+             ('operating_unit_ids', '=', self.main_OU.id)])
```
grindtildeath, line 67: Not sure this is working as operating_unit_ids is a x2m field. You probably should use `in` instead of `=` or remove the line all together as you're already searching with an id.

grindtildeath (changes requested):
@Matiar 
Thanks for your contribution. Here are a few remarks/recommandations.

hr_employee_operating_unit/security/security.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
```
grindtildeath, line 3: Missed in the last review but please add `noupdate="1"` here.

hr_employee_operating_unit/security/security.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    <record id="ir_rule_hr_emp_allowed_operating_units" model="ir.rule">
+        <field name="model_id" ref="hr.model_hr_employee"/>
+        <field name="domain_force">['|',('operating_unit_ids','=',False),
+                                        ('operating_unit_ids','in',[g.id for g in user.operating_unit_ids])]</field>
+        <field name="name">Employee operating units Allowed</field>
+        <field eval="1" name="perm_unlink"/>
+        <field eval="1" name="perm_write"/>
+        <field eval="1" name="perm_read"/>
+        <field eval="1" name="perm_create"/>
```
grindtildeath, line 13: I'm not sure about this but all the other modules only define perm_read to 1 on the record rule. Is there any reason to set all the perms to 1 ?

AaronHForgeFlow:
Hi @Matiar are you still working on this?

PR #99, opened 2017-11-03
[11.0] mig operating unit

gdgellatly:
- Change groups_id inheritance in view to specific group as groups_id gets replaced by a same priority view.
- Remove encodings
- Bump Version
- Test
- Add known issues for docs - if you know the answer I can update docs.
- Update name_search signature to match BaseModel

PR #100, opened 2017-11-03
11.0 mig sale team operating unit

gdgellatly:
Depends on

- [x] #99 operating_unit

Just needed a view fix as more_info page is now gone. Put with company as per v10.


AaronHForgeFlow:
Why are you including the operating_unit module fields here? #99  means to do that, doesn't' it?

gdgellatly:
@aheficent That is correct, this PR extends the last one, they include the same commits so they won't duplicate, they are all mergeable, but need to merge in order.  I don't really know how to explain it, but its right.  Maybe @pedrobaeza can explain better.

gdgellatly:
@aheficent To review you are best to just take the last commit, https://github.com/OCA/operating-unit/pull/100/commits/10f25f44133554bbf9c2a1754c91f116b2e1839b that is just the changes from v10 to v11 for sale_team

AaronHForgeFlow:
It's ok then. I'd preferred to do that using the oca_dependencies.txt file because now I get some merge conflicts when merging the two branches.

gdgellatly:
Hmmm, I think I made a mistake then, I'll look on Monday, but I did a merge commit when I probably shouldn't have.  Was my first time having to follow this process for multiple dependent modules.

AaronHForgeFlow:
np @gdgellatly as runbot is not working for v11 and I have to test in local anyway it's ok as it is

AaronHForgeFlow (commented):
IMHO I should not be able to add user to a sales team with  an operating unit when the user it's not assigned to that operating unit.

pedrobaeza:
About including commits of dependent modules: this is another technique not specified in the migration guide. What this does is to allow to test the module in this PR without waiting the dependency to be merged, but has these disadvantages:

* Confuse the commit list
* If the dependency PR evolves, this PR doesn't reflect it unless you make a git operation. In this case, the best is to rebase over dependency branch, instead of merging, or the history will be totally messed.
* When the other PR is merged, and more if a squashing operation is performed before merging, you have to make an interactive rebasing on this PR, removing manually commits from the other PR, or you will get a conflict.

gdgellatly:
@aheficent As I say, these are straight ports.  So the functionality has not changed.  I can kind of see both cases, but if people agree I'll add the domain.  It's not hard.

gdgellatly:
I'll probably just create a new branch and cherry pick as we merge up and then force push as we merge each PR.  Thats why I'm only doing a few modules at a time in dependent order.  In the meantime at least we can get reviews sorted.

AaronHForgeFlow:
@gdgellatly It's ok to me to keep the functionality as it was in v10. Better to create a PR with the fix in v10 and forward port it to v11 after.

operating_unit/README.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Usage
+=====
+
+This module defines the operating unit entity and the user's security rules.
+Other modules extend the standard Odoo apps with the OU.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/213/10.0
```
simahawk: /11.0

operating_unit/__init__.py
```diff
@@ -0,0 +1,5 @@
+# © 2015-2017 Eficent
```
simahawk: `Copyright`

operating_unit/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+
+{
+    "name": "Operating Unit",
+    "summary": "An operating unit (OU) is an organizational entity part of a "
+               "company",
+    "version": "11.0.1.0.0",
+    "author": "Eficent, "
+              "Serpent Consulting Services Pvt. Ltd.,"
+              "Odoo Community Association (OCA)",
+    "website": "http://www.eficent.com",
```
simahawk: https://github.com/OCA/operating-unit

simahawk:
@gdgellatly can you drop #99 commits and use `oca_dependencies.txt`? I know it makes your life easier in the short term but it gets harder for you and reviewers in the long one ;)

simahawk:
@gdgellatly #99 got merged: you can rebase :wink:

sales_team_operating_unit/security/crm_security.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016-17 Eficent Business and IT Consulting Services S.L.
+     Copyright 2017-TODAY Serpent Consulting Services Pvt. Ltd.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0) -->
+<odoo noupdate="0">
```
grindtildeath: `noupdate` should be 1

grindtildeath (commented):
Just a small change needed.

gdgellatly:
@simahawk it was already rebased and updated to work with changes in #99

sales_team_operating_unit/tests/test_crm_team_operating_unit.py
```diff
@@ -0,0 +1,66 @@
...
+#   (<http://www.serpentcs.com>)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo.tests import common
+
+
+class TestSaleTeamOperatingUnit(common.TransactionCase):
+
+    def setUp(self):
+        super(TestSaleTeamOperatingUnit, self).setUp()
+        self.res_users_model = self.env['res.users']
```
simahawk: `.with_context(tracking_disable=True, no_reset_password=True)` to avoid mail.thread overhead and avoid email creation overhead (and they are actually sent, BTW :))

simahawk (changes requested):

sales_team_operating_unit/tests/test_crm_team_operating_unit.py
gdgellatly: @simahawk Most of those issues were long fixed as part of rebase.  Not sure how you even found them given the code was force pushed and overwritten.

simahawk (approved):

PR #105, opened 2017-11-22
[10.0][FIX] Purchase Operating Unit

AaronHForgeFlow:
Current behavior: Created procurement in customers. Created PO to satisfy this procurement. Constraint raising: `The PO OU is not the same OU as the procurement location`

Expected behavior: As the PO is supposed to deliver directly to customers, the constraint should not raise.

Solution: Avoid raising a constraint when the procurement location is not in the company.

Update: The constraint should check the condition only when there's a PO actually.

### DIFF
--- a/purchase_operating_unit/models/procurement.py
+++ b/purchase_operating_unit/models/procurement.py
@@ -15,8 +15,9 @@ class ProcurementOrder(models.Model):
     def _check_purchase_order_operating_unit(self):
         for proc in self:
             if proc.location_id.usage not in ('supplier', 'customer') and \
-                (proc.purchase_line_id.order_id.operating_unit_id !=
-                    proc.location_id.operating_unit_id):
+                proc.purchase_line_id.order_id and \
+                    (proc.purchase_line_id.order_id.operating_unit_id !=
+                        proc.location_id.operating_unit_id):
                 raise ValidationError(
                     _('Configuration error. The Quotation / Purchase Order '
                       'and the Procurement Order must belong to the same '
--- a/purchase_operating_unit/models/procurement.py
+++ b/purchase_operating_unit/models/procurement.py
@@ -10,18 +10,20 @@ from odoo.exceptions import ValidationError
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
 
-    @api.one
+    @api.multi
     @api.constrains('purchase_line_id')
     def _check_purchase_order_operating_unit(self):
-        purchase = self.purchase_line_id.order_id
-        location = self.location_id
-        if (purchase and
-                purchase.operating_unit_id != location.operating_unit_id):
-            raise ValidationError(
-                _('Configuration error\nThe Quotation / Purchase Order and '
-                  'the Procurement Order must belong to the same '
-                  'Operating Unit.')
-                )
+        for proc in self:
+            purchase_ou = proc.purchase_line_id.order_id.operating_unit_id
+            location_ou = proc.location_id.operating_unit_id
+            usage = proc.location_id.usage
+            if usage not in ('supplier', 'customer'):
+                if purchase_ou != location_ou:
+                    raise ValidationError(
+                        _('Configuration error. The Quotation / Purchase Order'
+                          ' and the Procurement Order must belong to the same '
+                          'Operating Unit.')
+                        )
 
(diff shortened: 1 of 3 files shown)


purchase_operating_unit/models/procurement.py
```diff
@@ -10,18 +10,20 @@
...
-                purchase.operating_unit_id != location.operating_unit_id):
-            raise ValidationError(
-                _('Configuration error\nThe Quotation / Purchase Order and '
-                  'the Procurement Order must belong to the same '
-                  'Operating Unit.')
-                )
+        for proc in self:
+            purchase_ou = proc.purchase_line_id.order_id.operating_unit_id
+            location_ou = proc.location_id.operating_unit_id
+            usage = proc.location_id.usage
```
gdgellatly: Strictly speaking unnecessary variable assignment - idk maybe you did for 80 char limit.

purchase_operating_unit/models/procurement.py
```diff
@@ -10,18 +10,20 @@
...
+            if usage not in ('supplier', 'customer'):
+                if purchase_ou != location_ou:
+                    raise ValidationError(
+                        _('Configuration error. The Quotation / Purchase Order'
```
gdgellatly: Convention is a space at end of string for continuation not start of following, as per original code.

purchase_operating_unit/models/procurement.py
```diff
@@ -10,18 +10,20 @@
 class ProcurementOrder(models.Model):
     _inherit = 'procurement.order'
 
-    @api.one
+    @api.multi
     @api.constrains('purchase_line_id')
```
gdgellatly: Question: I don't use v10 anymore, but doesn't the constraint now also apply to location_id?  I doubt it would trigger in practice anyway.  On second thoughts, that gets problematic when considering other modules that might raise procurement.

purchase_operating_unit/models/procurement.py
```diff
@@ -10,18 +10,20 @@
...
-                _('Configuration error\nThe Quotation / Purchase Order and '
-                  'the Procurement Order must belong to the same '
-                  'Operating Unit.')
-                )
+        for proc in self:
+            purchase_ou = proc.purchase_line_id.order_id.operating_unit_id
+            location_ou = proc.location_id.operating_unit_id
+            usage = proc.location_id.usage
+            if usage not in ('supplier', 'customer'):
+                if purchase_ou != location_ou:
```
gdgellatly: There is no point to this embedded if, the code is a single branch, so better as single if.  

The above three lines could be simply written as
```python
internal_usage = proc.location_id.usage not in ('supplier', 'customer')
if internal_usage and purchase_ou != location_ou:
```

But these are all minor points.  Code does what it says.  Its life is ending now anyway as no procurement model in v11 so not blocking comments.

gdgellatly (approved):
Minor comments.  None blocking, mostly matters of opinion. Bugfix fixes the bug.

AaronHForgeFlow:
@gdgellatly I agree with all comments. Thanks

LoisRForgeFlow (approved):
Code LGTM

MiquelRForgeFlow (approved):

purchase_operating_unit/models/procurement.py
```diff
@@ -10,18 +10,19 @@
...
-        location = self.location_id
-        if (purchase and
-                purchase.operating_unit_id != location.operating_unit_id):
-            raise ValidationError(
-                _('Configuration error\nThe Quotation / Purchase Order and '
-                  'the Procurement Order must belong to the same '
-                  'Operating Unit.')
-                )
+        for proc in self:
+            if proc.location_id.usage not in ('supplier', 'customer') and \
```
simahawk: I'd rewrite it like this to ease readability (not blocking tho):
```python
if not proc.purchase_line_id:  # guess that if you have a PO line you the order too, right? :)
    continue
ou = proc.operating_unit_id
order_ou = proc.purchase_line_id.order_id.operating_unit_id
if (ou != order_ou and
       proc.location_id.usage not in ('supplier', 'customer')):
[...]
```

simahawk (approved):

grindtildeath (approved):
Out of module version number to increase :+1:

PR #112, opened 2018-04-10
[FIX][10.0] stock_operating_unit 

AaronHForgeFlow:
The constrains in the method `_check_stock_move_operating_unit` are not correct. As a result we get in the logs: `@constrains parameter' operating_unit_id' is not writeable`


LoisRForgeFlow (approved):
Makes sense

HviorForgeFlow (approved):
:+1:

PR #115, opened 2018-05-15
[10.0][FIX]deprecated method

AaronHForgeFlow:
UPDATE: keep the method without api as the original one.

ORIGINAL SUMMARY:
If the method def `_get_new_picking_values` in stock_move is inherited in any other module in the way:
```
@api.model
def _get_new_picking_values(self, move):
```
this will crash.

### DIFF
--- a/sale_stock_operating_unit/models/stock_move.py
+++ b/sale_stock_operating_unit/models/stock_move.py
@@ -13,7 +13,7 @@ class StockMove(models.Model):
         """
         Override to add Operating Units to Picking.
         """
-        values = super(StockMove, self)._prepare_picking_assign()
+        values = super(StockMove, self)._get_new_picking_values()
         sale_line = self.procurement_id and self.procurement_id.sale_line_id
         if sale_line:
             values.update({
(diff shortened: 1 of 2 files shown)


sale_stock_operating_unit/models/stock_move.py
```diff
@@ -9,6 +9,7 @@
 class StockMove(models.Model):
     _inherit = 'stock.move'
 
+    @api.multi
```
MiquelRForgeFlow: why not `api.model`?

AaronHForgeFlow: Because the original method is using a record in self:

https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_move.py#L463

Is it better to call https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_move.py#L451 directly?

MiquelRForgeFlow: Delete first commit

MiquelRForgeFlow (approved):

AaronHForgeFlow:
I will close this because the original message was wrong and it will be confusing. I will create a new one including this fix. Sorry.

Issue #93, opened 2017-09-29
Error When i Try to Create New Quotation


aitlamsali:
Hi;

When i attach a user to his own UO and i connect with hiw account to make a quotation i got this error :

Odoo Warning - Access Error
The requested operation cannot be completed due to security restrictions. Please contact your system administrator.

(Document type: crm.team, Operation: read)

aitlamsali:
Hi;

When i attach a user to his own UO and i connect with hiw account to make a quotation i got this error :

Odoo Warning - Access Error
The requested operation cannot be completed due to security restrictions. Please contact your system administrator.

(Document type: crm.team, Operation: read)

aitlamsali:
Hello could you help me please ?

AaronHForgeFlow:
Hi @aitlamsali Thank you for your report. Which Odoo version are you using. It seems to me that user should belong to group `Sales / User: Own Documents Only` Does the user belong to that group or Sales Manager group?

w-mohsen:
Hi @aheficent I have the same issue in Odoo 10 and I got the same error when the user belong to Sales / User: Own Documents Only group and when I change it to User: All Documents group, I was able to create quotation / Sale orders.

 Please advise.

AaronHForgeFlow:
@aitlamsali  @w-mohsen Thank you for reporting this issue. Can you check this PR https://github.com/OCA/operating-unit/pull/116? Thanks

PR #116, opened 2018-05-16
[10.0][FIX] sale_operating_unit

AaronHForgeFlow:
Basic sales user is not able to create SO because the code is searching on sales_team

It solves this issue https://github.com/OCA/operating-unit/issues/93

### DIFF
--- a/sale_operating_unit/models/sale.py
+++ b/sale_operating_unit/models/sale.py
@@ -30,11 +30,11 @@ class SaleOrder(models.Model):
     @api.onchange('team_id')
     def onchange_team_id(self):
         if self.team_id:
-            self.operating_unit_id = self.team_id.operating_unit_id
+            self.operating_unit_id = self.sudo().team_id.operating_unit_id
 
     @api.onchange('operating_unit_id')
     def onchange_operating_unit_id(self):
-        if self.team_id and self.team_id.operating_unit_id != \
+        if self.team_id and self.sudo().team_id.operating_unit_id != \
                 self.operating_unit_id:
             self.team_id = False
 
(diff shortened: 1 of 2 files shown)


sale_operating_unit/models/sale.py
```diff
@@ -12,7 +12,7 @@ class SaleOrder(models.Model):
 
     @api.model
     def _default_operating_unit(self):
-        team = self.env['crm.team']._get_default_team_id()
+        team = self.env['crm.team'].sudo()._get_default_team_id()
```
MiquelRForgeFlow: This sudo makes absolutely sense to me. The rest, no.

AaronHForgeFlow: what about self.team_id.sudo() ?
If we remove the sudos completely users will get an AccessError

MiquelRForgeFlow: why? the sudo is for accessing methods, not values. sudo is for browse, search, etc.

AaronHForgeFlow: I don't know :see_no_evil:  

The thing is I got AccessError if don't use that. I don't see any other solution.

MiquelRForgeFlow: Could you dive a bit deeper in the code while debugging to see the cause? :)

AaronHForgeFlow: The problem is the search in the sales team. Either we search using sudo, or we give permissions to the user.

MiquelRForgeFlow: I prefer the sudo in the search

AaronHForgeFlow:
o-no-idea-dog-meme-facebook

AaronHForgeFlow:
This issue is not actually happen or at least I cannot reproduce it. Closing

PR #120, opened 2018-06-07
[10.0][FIX]stock_operating_unit

AaronHForgeFlow:
Currently we cannot assign an operating unit to a locations that is not 'internal'. IMHO this is wrong. We should add the operating unit to locations type view, production or every usage different than customers/suppliers.


AaronHForgeFlow:
@Eficent

LoisRForgeFlow (approved):
I agree :+1:

MiquelRForgeFlow (approved):

PR #137, opened 2018-07-27
Replaces #114 - Squashed commit of the following:

richard-willdooit:
commit 42ebcc6efc35989249a3403c59f15201bde8a856
Author: Richard deMeester <richard@willowit.com.au>
Date:   Fri Jul 27 10:39:07 2018 +1000

    Remove erroneous extra currency code.

commit 921aa2c9bf87ff920bfa3debfd1b3ef534d059a5
Author: Michael Aldrin Villamar <michael.villamar@cloudemployee.co.uk>
Date:   Thu Jul 26 16:04:32 2018 +0800

    [FIX][account_payment] Fix in the internal transfers amount currency; [FIX] Allow selecting OU on all account journals

commit b6ce928dbcdd9dd706dcd86bb78babe7aeb70f88
Merge: e071cb4 b9c083c
Author: Michael Aldrin Villamar <michael.villamar@cloudemployee.co.uk>
Date:   Thu Jul 26 15:25:55 2018 +0800

    Merge branch '11.0-willdoo_account_operating_unit_module_upgrade' of github.com:WilldooIT/operating-unit into 11.0-willdoo_account_operating_unit_module_upgrade

commit b9c083ce807f0e687f40afa32dd1064a8ee9e2ab
Author: Michael Aldrin Villamar <michael.villamar@cloudemployee.co.uk>
Date:   Tue Jul 17 12:55:42 2018 +0200

    [MIG]account_operating_unit to v11

commit fd0433ec2a5eea9ac37902e53c0d37a7ccaf0862
Author: Isabel Esparza <isabel.esparza@jarsa.com.mx>
Date:   Tue Apr 10 12:20:52 2018 -0500

    [FIX][account_payment]Fix in the internal transfers amount currency

commit aa18228b9c6e7c7c329ff6d951bb08f784129bfd
Author: aaron <ahenriquez@eficent.com>
Date:   Mon Jun 11 12:02:48 2018 +0200

    [FIX]PYLINT mainly

commit cb65d729ec6c7736d1f14255b1185fc432f91a27
Author: Stéphane Bidoul (ACSONE) <stephane.bidoul@acsone.eu>
Date:   Wed May 16 14:15:40 2018 +0200

    Allow selecting OU on all account journals

    Closes #118

commit 368b1fb5ea18bf5a42ea9a5b0b245fd9df55a2ba
Author: Darshan Patel <darshan.patel.serpentcs@gmail.com>
Date:   Tue Jul 4 14:48:56 2017 +0530

[...]

PR #114, opened 2018-04-16
[MIG] Migrate account_operating_unit to 11.0

michael-villamar:

PR #140, opened 2018-10-03
[11.0][MIG] account_operating_unit : Migration to v11.0

grindtildeath:
Supersedes : #114 and #137 

 - I did remove the first commits on #114 (adding both account_operating_unit and analytic_operating unit) which are reverted further in the PR, to keep a clean commit history.
 - I did extract the change explained by @richard-willdooit in a separated commit and did copy his comment from https://github.com/OCA/operating-unit/commit/fd2fc49de45f64774bf14f7cb7df026d5006f24a to have a better explanation of the change on the commit itself, so it can be backported if needed.
 - I finally did the changes asked by @nikul-serpentcs on #114 in a separated commit.

@michael-villamar @meigallodixital Since it seems both of you did the job, I hope this is ok for you, and thank you BTW.


AaronHForgeFlow (approved):
:+1: Tested on local.

Reviewed the interOU invoice / payment flow including correct accounting entries. Reviewed BS and P&L by operating unit.

Thanks also for fixing the commit history mess.

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,146 @@
...
+from odoo.exceptions import UserError
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    operating_unit_id = fields.Many2one('operating.unit', 'Operating Unit',
+                                        default=lambda self:
+                                        self.env['res.users'].
+                                        operating_unit_default_get(self._uid))
```
grindtildeath: @aheficent As I didn't do the migration, I might just ask here, but this is getting me into troubles on https://github.com/OCA/account-consolidation/pull/16

As you can see there, we try to use operating_unit in a multicompany context, and this is now breaking travis with a infinite recursion between https://github.com/OCA/operating-unit/blob/11.0/operating_unit/models/res_users.py#L17 and its defaut value https://github.com/OCA/operating-unit/blob/11.0/operating_unit/models/res_users.py#L31.

So could we get rid of this default value here ? 

IMO we'd also have to migrate `default_operating_unit_id` on `res.users` to a property field if we want to support multicompany. But at the moment it would ease my job on the consolidation module to remove the default here.

Or maybe you have another solution from where you use these modules ?

AaronHForgeFlow: I see that the default operating unit should be company dependent but that error is strange. I'll check the consequences of removing the default here. ping @jbeficent @mreficent

MiquelRForgeFlow: Odoo 11.0 + multicompany + operating_unit = 🤯

AaronHForgeFlow: @grindtildeath can you provide the steps to reproduce the issue?

grindtildeath: @aheficent I tried to reproduce the issue on UI but it seems I can't reproduce the recursion issue as it happens on travis... Anyway we'll probably abandon the use of operating.unit in the context of multicompany so we will probably close my PR https://github.com/OCA/account-consolidation/pull/16 .

However, as I see it, the default value as implemented here doesn't respect this help here : https://github.com/OCA/operating-unit/pull/140/files#diff-14450c55bf5d67906de8a1d52d658a9dR61

So I guess either the help or the default value should be corrected ?

AaronHForgeFlow: Correct. The default value in the journal items should be the one in the journal entry for manual journal entries. It is a bug.

meigallodixital:
Hello everyone, I'm alive! ;)

I was busy with other projects, but I think this week I was definitely free to dedicate myself completely to odoo. I will test these changes and comment. :)

As I see, it has been cleaned again and mix the changes.

meigallodixital:
What is still to be revised and / or done so that the changes are accepted?

AaronHForgeFlow:
@meigallodixital It will be nice if contributors make some functional reviews here, there are many features added by this module. I tested and it was fine to me but maybe I missed something.

meigallodixital:
> 
> 
> @meigallodixital It will be nice if contributors make some functional reviews here, there are many features added by this module. I tested and it was fine to me but maybe I missed something.

For us it works correctly, in fact we are already testing the operating unit on purchases and sales.

We are still assembling the erp skeleton, so we do not have a complete accounting, but in the tests we have not noticed any error.

AaronHForgeFlow:
@meigallodixital  that's great. It'd be great if you approve the Pull request then, thus everybody will see that you have reviewed.

meigallodixital (approved):
For us it works correctly, in fact we are already testing the operating unit on purchases and sales. We have not noticed any error.

AaronHForgeFlow:
I tested again and It looks good to me. As nobody has detected any issue I am merging this.

MiquelRForgeFlow (approved):

PR #142, opened 2018-10-17
[10.0][FIX]wrong constraint on hr_payroll_account_operating_unit

AaronHForgeFlow:
The constraint was not useful as it will never show an error.
Added related field, it makes sense as the operating unit in the payslip is inherited form the contract.
Also, the operating unit was not propagated correctly.


bjeficent (approved):
LGTM

AdriaGForgeFlow (approved):
Code review :+1:

PR #144, opened 2018-11-19
[12.0][MIG] operating_unit

AdriaGForgeFlow:
Migration of operating_unit to v12.0

PR #145, opened 2018-11-20
[12.0][MIG] sales_team_operating_unit

bjeficent:
Migration of sales_team_operating_unit to version 12.0


nikul-serpentcs:
@bjeficent here remove `Operating Unit` module, because it's migrated in #144 

Check with @aheficent

PR #146, opened 2018-11-21
[12.0][MIG] Migrated analytic_operating_unit to v12

sudhir-erpharbor:
Migrated analytic_operating_unit to v12.
@jbeficent Please review and provide feedback.

PR #148, opened 2018-11-30
[12.0] [FIX] operating unit security

AaronHForgeFlow:
When creating sales_team (only for creating not editing) the OU manager can see all the operating units even if the manager is not assigned to those. Selecting an OU the manager is not assigned cause an access error when creating the sales team.

Despite the above use case is a clear configuration problem (The OU manager should belong to all OU), we should prevent this error to be raised. IMO the most elegant way to do that is through domains in the fields as it is proposed for sales_team here.

 Notice that this domain is just preventing the manager to select any OU in the sales team. The basic user "Multi operating unit user" currently **cannot see any OU he/she is not assigned to**. This is only for the manager.

Please also notice that the new field in the operating unit will not create any table as it is already created by the field operating_unit_ids in res.users.

@sudhir-erpharbor If you think this is ok to you I will apply the same domain for analytic accounts at #146

@bjeficent  If this is accepted we will apply the same thing to the module account_operating_unit

### DIFF
--- a/operating_unit/models/operating_unit.py
+++ b/operating_unit/models/operating_unit.py
@@ -17,6 +17,10 @@ class OperatingUnit(models.Model):
         'res.company', 'Company', required=True, default=lambda self:
         self.env['res.company']._company_default_get('account.account'))
     partner_id = fields.Many2one('res.partner', 'Partner', required=True)
+    user_ids = fields.Many2many(
+        'res.users', 'operating_unit_users_rel', 'poid', 'user_id',
+        'Users Allowed',
+    )
 
     _sql_constraints = [
         ('code_company_uniq', 'unique (code,company_id)',
--- a/operating_unit/README.rst
+++ b/operating_unit/README.rst
@@ -57,6 +57,13 @@ Usage
 This module defines the operating unit entity and the user's security rules.
 Other modules extend the standard Odoo apps with the OU.
 
+Known issues / Roadmap
+======================
+
+The column name of the table operating_unit_users_rel is bad. Change it in v13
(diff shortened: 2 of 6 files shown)


AaronHForgeFlow:
@gdgellatly you also may be interested on this

operating_unit/models/operating_unit.py
```diff
@@ -17,6 +17,10 @@ class OperatingUnit(models.Model):
         'res.company', 'Company', required=True, default=lambda self:
         self.env['res.company']._company_default_get('account.account'))
     partner_id = fields.Many2one('res.partner', 'Partner', required=True)
+    user_ids = fields.Many2many(
+        'res.users', 'operating_unit_users_rel', 'poid', 'user_id',
```
MiquelRForgeFlow, line 21: poid?

AdriaGForgeFlow (approved):
LGTM :+1:

operating_unit/models/operating_unit.py
AaronHForgeFlow: Yes, that was the column of the original field. But I agree that is not correct and we should change both. I'll do the thing.

AaronHForgeFlow: Actaully, as the module is merged, I think it is better to do the thing in v13. I'll update the readme instead.

nikul-serpentcs (approved):
Code Review LGTM

AaronHForgeFlow:
@mreficent is this fine to you?

MiquelRForgeFlow (approved):

PR #158, opened 2019-01-21
[12.0] [FIX] operating_unit

bjeficent:
The creation of Operating Units were failing cause of wrong environment.


AaronHForgeFlow (approved):
:+1: Fast tracking this as it is a clear bug.

PR #150, opened 2018-12-12
[12.0] [MIG] stock_operating_unit

bjeficent:
Migration of module stock_operating_unit to version 12.0

PR #149, opened 2018-12-10
[12.0][MIG] account_operating_unit

bjeficent:
Migration of module to version 12.0

PR #162, opened 2019-01-29
[12.0][FIX]clear caches when saving so OU rules are updated

AaronHForgeFlow:
Access rules are not updated right away because of the cache. With this change we prevent the system to raise Access Error on newly created OU. The user that creates the OU is auto-assign to the OU but the rule is not updated caused of the bad cache.  This fixes the issue.

### DIFF
--- a/operating_unit/__manifest__.py
+++ b/operating_unit/__manifest__.py
@@ -6,7 +6,7 @@
     "name": "Operating Unit",
     "summary": "An operating unit (OU) is an organizational entity part of a "
                "company",
-    "version": "12.0.1.1.0",
+    "version": "12.0.1.2.0",
     "author": "Eficent, "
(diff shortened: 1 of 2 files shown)


bjeficent (approved):
Tested with #149 and #150, works like a charm :+1:

bjeficent:
I have updated this to avoid crashes in multi-company environments. The company of the OU's were editable and when you changed to another company an access error was prompted. This fix makes the company readonly and takes the current company to the field. If you want to create OU's with different companies, you have to change the company first.

AdriaGForgeFlow (approved):
LGTM :+1:

AaronHForgeFlow:
Merging as it is blocking other PRs and CI is good. Tested again and it works good.

PR #165, opened 2019-02-01
[11.0] [FIX] account_operating_unit: Consistency between operating units

bjeficent:
The operating units of the Invoice and the Journal must be the same.
Files renamed for better organization


elicoidal (approved):
LGTM
I would suggest to bump the version and to add a test

AaronHForgeFlow (approved):
:+1:

PR #176, opened 2019-02-15
[11.0] [FIX] Counterpart's operating unit is wrong and Journals filtered by OUs

bjeficent:
The counterpart calculated in Internal Transfers had a wrong OU. The Journals were not filtered by OU's

### DIFF
--- a/account_operating_unit/models/account_invoice.py
+++ b/account_operating_unit/models/account_invoice.py
@@ -18,6 +18,29 @@ class AccountInvoice(models.Model):
                                         states={'draft': [('readonly',
                                                            False)]})
 
+    @api.onchange('operating_unit_id')
+    def _onchange_operating_unit(self):
+        if self.operating_unit_id and (
+                not self.journal_id or
+                self.journal_id.operating_unit_id != self.operating_unit_id):
+            journal = self.env['account.journal'].search(
+                [('type', '=', self.journal_id.type)])
+            jf = journal.filtered(
+                lambda aj: aj.operating_unit_id == self.operating_unit_id)
+            if not jf:
+                self.journal_id = journal[0]
+            else:
+                self.journal_id = jf[0]
+
+    @api.onchange('journal_id')
+    def _onchange_journal(self):
+        if self.journal_id and self.journal_id.operating_unit_id and \
+                self.journal_id.operating_unit_id != self.operating_unit_id:
+            ou = self.env['operating.unit'].search(
+                [('id', '=', self.journal_id.operating_unit_id.id)], limit=1
+            )
+            self.operating_unit_id = ou
+
     @api.multi
     def finalize_invoice_move_lines(self, move_lines):
         move_lines = super(AccountInvoice,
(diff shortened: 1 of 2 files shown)


account_operating_unit/models/account_invoice.py
```diff
@@ -18,6 +18,26 @@ class AccountInvoice(models.Model):
...
+            rf = res.filtered(
+                    lambda aj: aj.operating_unit_id == self.operating_unit_id)
+            if not rf:
+                self.journal_id = res[0]
+            else:
+                self.journal_id = rf[0]
+
+    @api.onchange('journal_id')
+    def _onchange_journal(self):
+        if self.journal_id:
```
JordiBForgeFlow: `if self.journal_id and self.journal_id.operating_unit_id and self.journal_id.operating_unit_id != self.operating_unit_id'

account_operating_unit/models/account_invoice.py
```diff
@@ -18,6 +18,26 @@ class AccountInvoice(models.Model):
...
+            res = self.env['operating.unit'].search(
+                [('id', '=', self.journal_id.operating_unit_id.id)]
```
JordiBForgeFlow: add `limit=1`

account_operating_unit/models/account_invoice.py
```diff
@@ -18,6 +18,26 @@ class AccountInvoice(models.Model):
                                         states={'draft': [('readonly',
                                                            False)]})
 
+    @api.onchange('operating_unit_id')
+    def _onchange_operating_unit(self):
+        if self.operating_unit_id:
```
JordiBForgeFlow: `if self.operating_unit_id and (not self.journal_id or self.journal_id.operation_unit_id != self.operating_unit_id`

account_operating_unit/models/account_invoice.py
```diff
@@ -18,6 +18,26 @@ class AccountInvoice(models.Model):
...
+            res = self.env['account.journal'].search(
+                [('type', '=', self.journal_id.type)])
+            rf = res.filtered(
+                    lambda aj: aj.operating_unit_id == self.operating_unit_id)
```
JordiBForgeFlow: Please change to `
journal = self.env['account.journal'].search([('type', '=', self.journal_id.type), ('operating_unit_id', '=', self.operating_unit_id.id), limit=1]
self.journal_id = journal
`

JordiBForgeFlow (changes requested):

account_operating_unit/models/account_invoice.py
AaronHForgeFlow: But if there is no specific journal for the OU, we can take a journal that is not assigned to any OU.

bjeficent: As @aheficent said, that way we could not take a journal without an assigned OU.

bjeficent:
Comments attended, thanks for reviewing.

AaronHForgeFlow:
Needs rebasing

bjeficent:
Rebase done

PR #178, opened 2019-02-21
[10.0] [FIX] stock_operating_unit

bjeficent:
Update constraint to verify that the picking type has a warehouse assigned. This error happens when the picking type is Dropship, because it has not a warehouse assigned by default.

### DIFF
--- a/stock_operating_unit/model/stock.py
+++ b/stock_operating_unit/model/stock.py
@@ -138,7 +138,7 @@ class StockPicking(models.Model):
     def _check_picking_type_operating_unit(self):
         for rec in self:
             warehouse = rec.picking_type_id.warehouse_id
-            if (rec.picking_type_id and rec.operating_unit_id and
+            if (rec.picking_type_id and rec.operating_unit_id and warehouse and
                     warehouse.operating_unit_id != rec.operating_unit_id):
                 raise UserError(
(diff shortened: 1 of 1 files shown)


elicoidal (approved):
LGTM

AaronHForgeFlow (approved):
:+1: LGTM. Pending to fix travis

AaronHForgeFlow:
Travis error is not related to this PR

AaronHForgeFlow:
@bjeficent can you rebase? Thanks

bjeficent:
Rebase done

stock_operating_unit/model/stock.py
```diff
@@ -139,6 +139,7 @@ def _check_picking_type_operating_unit(self):
         for rec in self:
             warehouse = rec.picking_type_id.warehouse_id
             if (rec.picking_type_id and rec.operating_unit_id and
+                    rec.picking_type_id.warehouse_id.operating_unit_id and
```
AaronHForgeFlow: Looking at this again, it seems to me that the chech should be like this:

`and warehouse`

 as there is a variable for that and  we are just checking if the warehouse is set in the picking type

bjeficent: You are right, changes done

PR #179, opened 2019-02-22
[10.0] [FIX] purchase_request_procurement_operating_unit: Wrong OU

bjeficent:
The OU is not propagated correctly when the picking type is Dropship.


MiquelRForgeFlow (approved):
LGTM

AaronHForgeFlow (approved):
:+1: Tests good so far

PR #180, opened 2019-02-25
[10.0] [FIX] hr_expense_operating_unit: Test failing

bjeficent:
The test of this module is failing in other PRs because of the exception type.


AaronHForgeFlow (approved):
Tested on local. Some action in test make two different constraints to raise. This unifies the error type.

LoisRForgeFlow (approved):

PR #187, opened 2019-03-20
[IMP] account_operating_unit: Use the user OU if the journal doesn't …

gustavovalverde:
…have it

### DIFF
--- a/account_operating_unit/models/account_payment.py
+++ b/account_operating_unit/models/account_payment.py
@@ -13,7 +13,7 @@ class AccountPayment(models.Model):
         for payment in self:
             if payment.journal_id:
                 payment.operating_unit_id = \
-                    payment.journal_id.operating_unit_id
+                    payment.journal_id.operating_unit_id or self.env.user.default_operating_unit_id or False
 
     operating_unit_id = fields.Many2one(
         'operating.unit', string='Operating Unit',
--- a/account_operating_unit/models/account_payment.py
+++ b/account_operating_unit/models/account_payment.py
@@ -13,7 +13,8 @@ class AccountPayment(models.Model):
         for payment in self:
             if payment.journal_id:
                 payment.operating_unit_id = \
-                    payment.journal_id.operating_unit_id or self.env.user.default_operating_unit_id or False
+                    payment.journal_id.operating_unit_id or \
+                    self.env.user.default_operating_unit_id or False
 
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
I am not sure. If we have different bank journals for different operating units then all of them should have an operating unit assigned. Maybe it is better to add a constraint to accomplish that, but not sure also.

gustavovalverde:
@aheficent There are 2 scenarios, both supported by this PR:
1- One journal to be used by different operating units
2- A journal per operating unit


For both use cases, the operating unit used by the payment is the one set in the journal; if the journal doesn't have it, then use the one in the user.

Without the PR, use case 1 can't be accomplished. With the PR I'm not seeing how use case 2 gets affected.

AaronHForgeFlow:
In case 1, which operating unit pays the bill? It needs to be one only one operating unit. Even if the operating unit is fully dedicated to pay the bills of the other operating units.

gustavovalverde:
Case 1, in a real life scenario:
#### Operating Units
- Administration
- Branch 1
- Branch 2
- Branch 3

### Journals
- Cash (no operating unit)
- Bank A (Branch 1)

All branches (OU) register their respective expenses. Financially, the cash used comes from the same source/account, so there's not need for multiple journals.

Each OU pays their bill, selecting the **Cash** journal. As the journal has no OU, the OU used on the payment is the one from the user registering the payment.

Maybe and additional validation could be that a payment from an OU, can't pay a bill from another.

AaronHForgeFlow:
I understand that case. Perhaps you are right. Any way I think it is a good idea to have three cash journals for every OU. 

An OU can pay bills from the other, for example, administration can pay the bills of all others. Then, if OUs are self balanced an extra journal items is created to balance.

But in any case, I think this PR will not affect that use case, so I will not block this.

gustavovalverde:
I signed the CLA, what else is needed so this PR can be merged? cc: @pedrobaeza

blaggacao:
#193 Should solve this quite elegantly. Journals are metadata. Period.

gustavovalverde:
ping @jbeficent @pedrobaeza

andres-pcg:
Any update on this one? @pedrobaeza

pedrobaeza:
I'm not involved in this repo, so you should ping another person

PR #189, opened 2019-04-07
[10.0][FIX]stock_Account

AaronHForgeFlow:
If no operating unit in the locations take it from the warehouse.

UPDATE: updated PR for a couple of extra fixes:
- post stock_journal inter-OU entry after creating it
- do not create entry if none of the locations involved has a company

### DIFF
--- a/stock_account_operating_unit/model/stock_quant.py
+++ b/stock_account_operating_unit/model/stock_quant.py
@@ -3,7 +3,7 @@
 # - Jordi Ballester Alomar
 # © 2015-17 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
-from openerp import api, models
+from odoo import api, models
 
 
 class StockQuant(models.Model):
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
forever_alone

stock_account_operating_unit/model/stock_quant.py
```diff
@@ -22,11 +22,11 @@ def _account_entry_move(self, move):
         if move.product_id.valuation == 'real_time':
             # Inter-operating unit moves do not accept to
             # from/to non-internal location
-            if (
+            if ((move.location_id.company_id or
+                 move.location_dest_id.company_id) and
```
MiquelRForgeFlow: Redundant comparison. Make it simple:

```python
              if (move.location_id.company_id and
              move.location_id.company_id ==
                    move.location_dest_id.company_id and
                    move.operating_unit_id != move.operating_unit_dest_id):
```

AaronHForgeFlow: Ok, It is fine, but it is as complicated as I did :smile:

MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

JoanSForgeFlow (approved):

PR #190, opened 2019-04-09
[10.0][FIX]account_operating_unit not working when reconciling

AaronHForgeFlow:
The balancing inter OU entries should be created qhen reconciling entries from different OU. This is not happening in current version. 

Besides, refunds are not assigned to the correct operating unit.

Moreover, write off moves should use the operating unit form the journal

Fixes here should be ported to next versions because it affects normal flow.

### DIFF
--- a/account_operating_unit/models/account_move.py
+++ b/account_operating_unit/models/account_move.py
@@ -51,6 +51,67 @@ class AccountMoveLine(models.Model):
                                   ' the Move Line and in the Move must be the'
                                   ' same.'))
 
+    @api.multi
+    def _prepare_writeoff_first_line_values(self, values):
+        res = super(
+            AccountMoveLine, self)._prepare_writeoff_first_line_values(values)
+        if not res.get('operating_unit_id', False):
+            res['operating_unit_id'] = self.operating_unit_id.id
+        return res
+
+    @api.multi
+    def _prepare_writeoff_second_line_values(self, values):
+        res = super(
+            AccountMoveLine, self)._prepare_writeoff_second_line_values(values)
+        if not res.get('operating_unit_id', False):
+            res['operating_unit_id'] = self.operating_unit_id.id
+        return res
+
+    @api.multi
+    def _prepare_inter_ou_balancing_partial_reconcile(
+            self, move, ou_id, debit, credit):
+        if not move.company_id.inter_ou_clearing_account_id:
+            raise UserError(_('Error!\nYou need to define an inter-operating\
+                unit clearing account in the company settings'))
+
+        res = {
+            'name': 'OU-Balancing',
+            'move_id': move.id,
+            'journal_id': move.journal_id.id,
+            'date': move.date,
+            'operating_unit_id': ou_id,
+            'account_id': move.company_id.inter_ou_clearing_account_id.id,
+            'debit': debit,
+            'credit': credit
+        }
+        return res
+
+    def _get_move_vals(self):
+        """ Return dict to create the payment move
+        """
+        journal = self.journal_id
+        name = self.name
+        return {
+            'name': name + 'OU Balance',
+            'date': self.date,
+            'ref': self.ref or '/',
+            'company_id': self.company_id.id,
+            'journal_id': journal.id,
+        }
+
+    @api.multi
+    def reconcile(self, writeoff_acc_id=False, writeoff_journal_id=False):
+        res = super(AccountMoveLine, self).reconcile(
+            writeoff_acc_id, writeoff_journal_id)
+        if (not len(self.mapped('operating_unit_id')) == 1 and
+                not self.env.context.get('reversal', False)):
+            partial = self.env['account.partial.reconcile'].search(
+                [('debit_move_id', 'in', self.ids),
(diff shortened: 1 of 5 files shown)


AaronHForgeFlow:
When paying a payment from an OU invoices from various OU, the inter OU balancing entries are not created. This PR also fixes that.

AaronHForgeFlow:
This will contain the main fix of this PR: https://github.com/Eficent/operating-unit/pull/87

MiquelRForgeFlow:
squash commits

account_operating_unit/models/account_move.py
```diff
@@ -51,6 +51,69 @@ def _check_move_operating_unit(self):
...
+    @api.multi
+    def reconcile(self, writeoff_acc_id=False, writeoff_journal_id=False):
+        res = super(AccountMoveLine, self).reconcile(
+            writeoff_acc_id, writeoff_journal_id)
+        if not len(self.mapped('operating_unit_id')) == 1:
+            partial = self.env['account.partial.reconcile'].search(
+                [('debit_move_id', 'in', self.ids),
+                 ('credit_move_id', 'in', self.ids),
+                 ('bal_move_id', '=', False)])
+            if not self.env.context.get('reversal', False):
```
daramousk: You could attach this conditional statement on line 108 and avoid the unnecessary `search` if both do not evaluate to `True`

account_operating_unit/models/account_move.py
```diff
@@ -144,3 +208,84 @@ def _check_ou(self):
...
+                    _prepare_inter_ou_balancing_partial_reconcile(
+                        move_id, ou_id, 0.0, self.amount)
+            else:
+                if not move_id:
+                    move_id = self.env['account.move'].create(
+                        self.debit_move_id._get_move_vals())
+                line_data = self.debit_move_id.\
+                    _prepare_inter_ou_balancing_partial_reconcile(
+                        move_id, ou_id, self.amount, 0.0)
+            if line_data:
```
daramousk: I believe there is no need for this conditional, `_prepare_inter_ou_balancing_partial_reconcile`, this will always be `True`.

account_operating_unit/models/account_move.py
```diff
@@ -144,3 +208,84 @@ def _check_ou(self):
...
+                amls.append(ml_obj.with_context(wip=True).
+                            create(line_data))
+        if amls:
```
daramousk: Same here, `amls` will always be `True`

account_operating_unit/models/account_move.py
```diff
@@ -144,3 +208,84 @@ def _check_ou(self):
...
+                rec.reverse_bal_entries()
+            if rec.debit_move_id.operating_unit_id != \
+                    rec.credit_move_id.operating_unit_id:
+                rec.bal_move_id = rec.create_ou_balance()
+
+    @api.multi
+    def unlink(self):
+        for rec in self:
+            if rec.bal_move_id and not self.env.context.get('reversal', False):
+                ctx = self.env.context.copy()
```
daramousk: If you want to add a value to the context without overriding other values that may exist, I believe there is no need for doing that `copy()`, you could just `rec.with_context(reversal=True).reverse_bal_entries()` as you have already done in https://github.com/OCA/operating-unit/pull/190/files#diff-14450c55bf5d67906de8a1d52d658a9dR265

All other instances of `context.copy()` should be adjusted.

account_operating_unit/models/account_payment.py
```diff
@@ -20,13 +20,21 @@ def _compute_operating_unit_id(self):
...
+            return res
         else:
-            res['operating_unit_id'] = self.operating_unit_id.id or False
-        return res
+            for inv in invoice:
+                res = super(AccountPayment,
+                            self)._get_counterpart_move_line_vals(
+                    invoice=inv)
+                res['operating_unit_id'] = inv.operating_unit_id.id or False
+                return res
```
daramousk: Is this over-indented? Here only the first invoice is taken into consideration. Also, what happens if in case we have multiple invoices, they have different `operating_unit_id`s?

AaronHForgeFlow: Actually this is not needed, everything should be done from partial reconcile and better keep always balance journal items in a separate entry.

Thanks for your review BTW, all comments were attended.

frlanxpertiz (commented):
in [727c432](https://github.com/OCA/operating-unit/pull/190/commits/727c432a7d0dde02ca3a394f36944ca9fde8d60e) why did you took the operating unit of the journal and not self.operating_unit_id ?

AaronHForgeFlow:
@frlanxpertiz I agree, better to take it from the move line. This needs to be changed.

AaronHForgeFlow:
> in [727c432](https://github.com/OCA/operating-unit/pull/190/commits/727c432a7d0dde02ca3a394f36944ca9fde8d60e) why did you took the operating unit of the journal and not self.operating_unit_id ?

@frlanxpertiz Sorry for the late update. This is not fixed. I will also port the fix to v13. It is wrong in that version

Issue #192, opened 2019-05-05
Advise


blaggacao:
@jbeficent Looks like I'm gonna join this effort and drop https://github.com/odoo/odoo/pull/19839/files

I might do kind of a intense sprint during the next few days while ensuring existing features of my solution and hope I can give some serious amount of valuable input. Hope stars are aligned and things can be merged quickly...

I'll be working on v12. Please ping the people who should take not of this and otherwise close, once acknowledged.

Anything I should know / take into account?

Best Regards,
David

JordiBForgeFlow:
you're welcome to join! We are already migrating to v12. Please start by reviewing all open PR's.

blaggacao:
There are some milestones in parallel, all in course. Are some outdated / abandoned? https://github.com/OCA/operating-unit/milestones

JordiBForgeFlow:
Yes. Focus on v12

blaggacao:
@jbeficent Would it be ~feasible~ thinkable to inherit `BaseModel`, detect on init if field `company_id` is present and inject `operating_unit_id` (+ generic business logic) in such case? I think this would somewhat reduce the need for maintaining so many modules / dependencies. (Same for the views, maybe on `js` or `qweb`).

```python
# Infer and adopt from company_id:
states={'draft': [('readonly', False)]})
```

```python
# Rather easy to generalize:
    @api.multi
    @api.constrains('operating_unit_id', 'company_id')
    def _check_company_operating_unit(self):
        for pr in self:
            if (
                pr.company_id and
                pr.operating_unit_id and
                pr.company_id != pr.operating_unit_id.company_id
            ):
                raise ValidationError(_('The Company in the Invoice and in '
                                        'Operating Unit must be the same.'))
        return True

    @api.multi
    @api.constrains('operating_unit_id', 'journal_id')
    def _check_journal_operating_unit(self):
        for ai in self:
            if (
                ai.journal_id.operating_unit_id and
                ai.operating_unit_id and
                ai.operating_unit_id != ai.journal_id.operating_unit_id
            ):
                raise ValidationError(_('The OU in the Invoice and in '
                                        'Journal must be the same.'))
        return True
```
We would be left with this type of business logic:
( I see how [this](https://github.com/odoo/odoo/issues/28549) would be extremely useful, in this case )
```python
    @api.multi
    def finalize_invoice_move_lines(self, move_lines):
        move_lines = super(AccountInvoice,
                           self).finalize_invoice_move_lines(move_lines)
[...]

PR #193, opened 2019-05-08
WIP generic operating units

blaggacao:
Ongoing concept work as per https://github.com/OCA/operating-unit/issues/192#issuecomment-489445452

- POC in WIP
- Generic operating unit implementation
- Sunset the need for maintaining one patch module for every supported module
- Rather provide model metadata within this module (It’s Better to Beg for Forgiveness than to Ask for Permission - just expose knowledge of some modules even if they won't be present in the environment)
- Implement "following transaction" business logic through `related` + `stored` *  **
- Make somewhat existing implicit OU realm implementation explicit and enforce it reliably
- Therefore classify models into transaction models and metadata models ***
- Adopt for a multiple selected operating unit user context (similar to new announced  Odoo's company implementation)

\* maintenance free strategy, more DRY
\** Usable for example in this scenario `account.journal` -> `account.move` -> `account.move.line`
\*** something that's missing from odoo from a long time to automate things :smile:

@jbeficent This is conceptually quite wicked, but I'd share so you get a sense where I'm aiming at. Maybe you can advise some pitfalls you know of that lay down that road I'm going.


operating_unit/models/operating_unit_mixins.py
```diff
@@ -0,0 +1,211 @@
...
+        Yet an existing transaction, can passively be "expulsed" of a realm if
+        one of it's linked metadata's operating units are modified and would
+        top sharing one with the transaction.
+
+        "Expulsed" transaction block on write, unless their realm boundary is
+        healed or they are made unbound.
+
+        TODO: Ensure through SQL queries that transactions that would be
+        explused are exposed with a warning + confirmation (with the possibility
+        for the user to make them unbound)
```
blaggacao, line 61: If that's not ensured there will be ugly leakages. And realms will be an incomplete design.

blaggacao (commented):
Important TODO

blaggacao:
I'm wondering what behavior should be desired on change of:
- `company_id` -> Reset OU to it's default? (strict company name spacing of OUs *)
- `user_id` ->This is a bit more tricky. Odoo itself has it's flaw here, I believe: If a general company secretary is creating a transactional document for a functional person, the realm should be adopted from that functional person (assigned at `user_id`) rather than from the secretary (from the environment). There was the `uid2` parameter for such case, but I rather believe it should be the default behavior (also in Odoo upstream, btw.).

to be discussed

\* strict company name spacing: Records that have no company (unboung / global) cannot have an OU, neither (also: unbound / global)

blaggacao:
For module overloading we have to consider two things:
- Specify model overloading with the mixins through a ~~post_module_load hook~~ (which seems not to exists, only post init semantics exist, which are concerned with database schema, though) - through simple defecting from odoo's model building in case the inherited model is not in the pool
- Always position operating_unit **last** in the module graph (so it can *add* itself in a standard `_inherit` mechanism)

blaggacao:
```python
class ModelBuildingEvictor(object):
    """ Evict Odoo's model building in a given registry if the inherited model
    is not present in that registry. Cooperative pattern with the
    module-is-last-in-graph-ensurer, to make sure all potential candidate models
    are present in the registry. """

    # It's a python classmethod, not an odoo.api.model! That means it works on
    # ModelBuildingEvictor or it's python inherited class in a pre-odoo-model
    # state of the world.
    @classmethod
    def _build_model(cls, pool, cr):
        parents = cls._inherit
        parents = [parents] if isinstance(parents, pycompat.string_types) else (parents or [])
        if any(model not in pool for model in parents):
            return
        super()._build_model(pool, cr)
```

The second point from above ("last in graph") might be a little trickier...

blaggacao:
Two things missing:
- Allow to create modules which depend on this (without creating circular dependencies). For example to expand operating_unit magic onto custom out-of-tree models.
- Implement a configurator for automated `ir.rule` and `ir.ui.view` manipulation.


(And of course the realm expulsion check on transaction data)

blaggacao:
There is one problem I overlooked, so far: 
- Parametric transitions from one independent transaction model to another (invoice creates move, but so does a picking). This might be the harder one.
- Odoo lacking context managers for Model Classes or proper standard model intercom interfaces, the thing I've come up with is to decorate functions that have been identified as downstreaming emitters.

Like so:

`action_validate = emit_operating_unit(action_validate)`

And catch it on the transactional model create method from the context:

```python
def create(self, vals):
	if not 'operating_units' in self.env.context or not self.env.context['operating_units]:
	    return super().create(vals)

	OUs = self.env.context['operating_units]
	assert isinstance(OUs, set), "emit_operating_unit should pass a RecordSet"  # Maybe the set tpe is not in the MRO?
	assert all(isinstance(ou, models.BaseModel) for ou in OUs), "operating units should be emitted as set of 
	if len(OUs) == 1:
	    vals['operating_unit_id'] = OUs[0].id
	    return super().create(vals)
	else:
	    if therading.current_thread().name.startswith('ir.cron'):
	        return super().create(vals)
            # Inreality: Launch a wizard / or dialog
	    print('Enter which Operating Unit to adopt (leave blank for creating an unbound transaction:')
	    ou_id = input()
	    vals['operating_unit_id'] = ou_id
	    return super().create(vals)
```

blaggacao:
Damn https://github.com/odoo/odoo/issues/28549

operating_unit/overloads.py
```diff
@@ -0,0 +1,54 @@
...
+class AccountTax(        models.Model, MBE): _inherit = [META,'account.journal'      ];
+# account.move -> account.move.line
+class AccountMove(       models.Model, MBE): _inherit = [ITX, 'account.move'         ];
+class AccountMoveLine(   models.Model, MBE): _inherit = [DTX, 'account.move.line'    ]; _ou_follows = 'move_id';
+# account.invoice -> account.invoice.line
+# account.invoice -> account.invoice.tax
+class AccountInvoice(    models.Model, MBE): _inherit = [ITX, 'account.invoice'      ];
+AccountInvoice.action_move_create = emit_operating_unit(lambda self: super(type(self), self).action_move_create)
+class AccountInvoiceLine(models.Model, MBE): _inherit = [DTX, 'account.invoice.line' ]; _ou_follows = 'invoice_id';
+class AccountInvoiceTax( models.Model, MBE): _inherit = [DTX, 'account.invoice.tax'  ]; _ou_follows = 'invoice_id';
```
blaggacao: This is really, really tight and clean. (and hopefully easy to maintain!)

blaggacao: This is really, really tight and clean. (and hopefully easy to maintain!)

blaggacao (commented):
Yeah!

blaggacao:
@jbeficent Could you summon some feedback _on the design choices_ from code savvy contributors to this repo?

blaggacao:
- Dropping defaults; `ir.default` is much more flexible and does cover the use case perfectly.
- Implement secretary logic (proxy through the `user_id` if this field exists)

blaggacao:
This is getting fun :-) Those factories and init overrides start to work :smiley:

operating_unit/overloads.py
```diff
@@ -0,0 +1,120 @@
...
+_meta('account.tax')
+_meta('account.journal')
+_inTX('account.move')
+_deTX('account.move.line', 'move_id')
+_inTX('account.invoice', ['action_move_create'])
+_deTX('account.invoice.line', 'invoice_id')
+_deTX('account.invoice.tax', 'invoice_id')
+
+# Is it true that ...
+_deTX('thirsty.camels.rock', 'invoice_ids')
```
blaggacao: Do they?

blaggacao (commented):
Question of the day... :stuck_out_tongue_winking_eye:

operating_unit/overloads.py
blaggacao: Explanation: This is just to check the `ModelBuildingEvictor` is working as expected...

blaggacao:
```
odoo_1           | 2019-05-12 18:11:34,552 1 INFO test89 odoo.modules.registry: module operating_unit: creating or updating database tables 
odoo_1           | 2019-05-12 18:11:35,096 1 INFO test89 odoo.models: Storing computed values of account.move.line.operating_unit_id 
odoo_1           | 2019-05-12 18:11:35,301 1 INFO test89 odoo.models: Storing computed values of account.invoice.line.operating_unit_id 
odoo_1           | 2019-05-12 18:11:35,334 1 INFO test89 odoo.models: Storing computed values of account.invoice.tax.operating_unit_id 
```

blaggacao:
To correct:
- Enforcing operating units at the model level (through new or old mechanism) implies every Employee user should also be Operatin Unit user. It just doesn't make sense to enforce on a model but then be able to exclude a group of users (not assigning them the "user" group)

This means the group condition on the views is to be dropped and the OU-user to be enforce on internal employees.
I guess this was a design inconsistency in the current implementation, or did I miss something?

blaggacao:
Yeeeha!
screenshot-localhost-2019 05 12-19-08-36

blaggacao:
Not too bad:
```
2019-05-13 03:13:21,151 1 INFO test89 odoo.modules.registry: module operating_unit: creating or updating database tables 
2019-05-13 03:13:22,859 1 INFO test89 odoo.models: Storing computed values of account.bank.statement.line.operating_unit_id 
2019-05-13 03:13:22,906 1 INFO test89 odoo.models: Storing computed values of account.voucher.line.operating_unit_id 
2019-05-13 03:13:23,048 1 INFO test89 odoo.models: Storing computed values of crossovered.budget.lines.operating_unit_id 
2019-05-13 03:13:23,255 1 INFO test89 odoo.models: Storing computed values of sale.order.line.operating_unit_id 
2019-05-13 03:13:23,319 1 INFO test89 odoo.models: Storing computed values of purchase.order.line.operating_unit_id 
2019-05-13 03:13:23,676 1 INFO test89 odoo.models: Storing computed values of hr.payslip.line.operating_unit_id 
2019-05-13 03:13:23,780 1 INFO test89 odoo.models: Storing computed values of stock.inventory.line.operating_unit_id 
2019-05-13 03:13:23,887 1 INFO test89 odoo.models: Storing computed values of stock.move.operating_unit_id
```

blaggacao:
@jbeficent @aheficent @lreficent @pedrobaeza @bjeficent @gdgellatly @grindtildeath @mreficent @sbidoul 
If you don't mind, I'd kindly ask to break this thing with your use cases. It's ready for some testing.
Instructions:
- `base_traversal_module` needs to go into `server_wide_modules`
- Install `operating_unit`
- Go to Models (in developer mode)
- Filter by Operating Unit Candidates
- Open one and toggle the operating unit switch (some might fail, those are still missing in `overloads.py`)

I've seen a transaction I didn't take into account like in `hr.expense` -> `hr.expense.sheet` where a independent transaction `hr.expense` becomes somewhat a dependent after assignation to `hr.expense.sheet`. I've to think that through a little more. Similar case with `stock.move` and `stock picking` in some occasions.

blaggacao:
After some feedback, I plan to work on docs and a couple of tests before polishing this off for (potential) acceptance.

operating_unit/models/operating_unit_mixins.py
blaggacao: Also check implication of potential metadata realm leakages.

operating_unit/overloads.py
blaggacao: It got better.

gdgellatly:
I'm really not a fan of this approach. It seems very overengineered for what essentially amounts to an extra field on a few objects. Although I'm also not a fan of the current approach either.

To do what this PR attempt in a simple, consistent and repeatable way, and after sometime in production with the existing modules, I'm starting to feel like the best approach would be to adapt `res.company` to enable some kind of sub company type, a child company with the parents accounting settings. That has issues in itself for shared items (e.g. bank accounts) and stock transactions but I feel they are more easily dealt with.

blaggacao:
@gdgellatly thank you for your feedback. The "overengineering" tries to advance discussion on some generic patterns, I've seen at the example / service of the operating unit, where I feel they are applicable. It's arguable indeed, that this engineering is "required" for OUs. But it seems to be "sufficient". I hope exposing / forming the idea of transversal realms inspires core one day.

As for the company hierachy,: I've been there. I feel it's not optimal either. Essentially with all those approaches we are shadowing something that already exists in certain shades, but not ultimately aligned for what we want:
- Accounting -> company (too double use)
- WMS -> Warehouse (good abstraction, actually, maybe another dimension not even needed)
- HR -> Departments (good abstraction)

When I prepared the list of all models that have a `company_id` set, I realized it's actually just some very few "realms" we are talking about. It got then clear to me that what we are actually doing is shadowing concepts (companys, departments, warehouses) for the special purpose of ensuring a "realm" traceability + access enforcement.

We COULD provide that traceability + access on every of those "main category" models (warehouses, etc...) in each "realm".

But I feel it's still useful to do so on an overall org level with traceability + access generics. Not only in the "accounting realm".

Maybe what would be interesting is a tighter integration of our "OU proxies" to each of those realms?

blaggacao:
@gdgellatly I digged it up. Here is where I'm coming from which seems going down the road you'd suggest: https://github.com/odoo/odoo/pull/19839/files
It only proxies Accounting. And does not give a way to isolate metadata data easily. It is a headache on intercompany warehouse transactions generating accounting entries (without having those intricate parts overridden).
An attempt to selectively share other metadata like `res.partner` is a headache too and done by partner tags (with translation issues to deal with).

While on the transaction level me previous comment might put boundaries of sense on implementations of TX OU realms (traceability), as for potentially overlapping metadata realms the `2many` approach has a strong conceptual argument.

gdgellatly:
@blaggacao I maintain my own production branch of operating units. It isn't perfect by any stretch, but it does address some complex issues. But even then it has workarounds. Take for example a payment from a customer that deals with multiple operating units, or a wage payment where it needs splitting over multiple OU's.  

Getting things like OU's into the bank rec widget is more than adding a field, in fact there is no field to add, at least in python. So our approach, rather than having to rewrite the reconciliation widget, ended up being a special analytic account type which converts to an OU.

Other complex cases, manufactured by one OU for delivery to another OU customer. 

But I bug fix this every day, in a big complex multicompany environment, it seems in different areas and thats why I can't see a generic approach working. 

Even things like setting a default OU for a payment - we need to identify what is being paid, but what if its left open? etc etc. Or warehouse locations, have no direct relation to warehouses, so we do that via stored computed field differently to other ou fields.

Or even simple things, like maybe you want partners available to all OU's, but the sales team is assigned to the partner, or the user for that matter. So it needs different security.

Or users who work across multiple OU's yet have only 1 default with no context switcher.

blaggacao:
@gdgellatly Thank's so much for your use cases!!!
Noted to burn some brain-joules on it:
- payment from a customer that deals with multiple operating units (I guess that's why I think partial reconcile shouldn't be of any OU, but I'll be thinking this through more properly)
- Splitting (aish a whole different transaction type, I'll try to get an abstract grasp on it)
- JavaScript widgets (I believe they still need their own module / individual implementation - splitting might potentially be an exception, though)
- Morphing transactions (manufactured by one OU for delivery to another OU customer) - I think that's already covered by this PR, actually. I intentionally din't enforce transaction-to-transaction realm, only metadata realms of a given transaction.
- default inference (payments) - I guess the issue is if it's external TX info fetched from an API. Because it's difficult to scope those exogenous information properly. For internal defaulting, you are basically saying that "ad personam" or "ad responsible" semantics is not sufficient in the case of payments. Is that correct? - For that to answer we have to put the "ad personam" semantics of OU in relation to unscoped exogenous information. At some point, they need scoping, if no other help available, that's gonna be manual. (It's like receiving unlabeled stock having to label them yourself)
-  Or warehouse locations, have no direct relation to warehouses - I would argue: both are just plain metadata, so it's not so hard, in this case.
- partner-salesteam: So it needs different security - The `ir.rule` that are created are not deleted (just made inactive), so customizing access in DB is relatively straight forward. (I might miss something of your argument here?)
- with no context switcher - Context switcher is on todo list. I'm just waiting for Odoo's implementation of multiple simultaneous companies in context and how they implement defaults to get inspired.

I guess you have very interesting use cases here, and I'd try not to surrender too quickly before them. I tried to understand them teleologically, but if you have the feeling I need additional info on any one of them, please shout out.

I guess there are solutions out there :-) We just have to think harder, better and clearer. (and with the right glossary)

blaggacao:
> payment from a customer that deals with multiple operating units

You might be playing on the payment invoice relation which is direct. But still it's transactional and there is no transaction-to-transaction realm enforcement, so a user which can see both realms can do the match (at least that's what's intended).

blaggacao:
Splitting and merging might be two of a kind. Both require user input, because there is no way to infer the intended target OU (exception: merging transactions coming all form the same OU). Splitting , though, requires a lot more metadata, like splitting ratios of some kind of value concept.

Probably the best thing for splitting we could do is to provide a hook machinery for a JS dialogue to let the user work on that metadata. Also, maybe we can come up with a list of typical splitting transitions, which might get us a clearer picture of the use cases...

- ~~Payment induced account.move~~ (already handled by link model: partial reconcile - which shall be unbound)
- Split Accounting Causation (invoice, asset depreciation, expenses etc) between the OUs. Probably in the accounting subsystem a relatively common problem.
- Split Stock pickings? (or prevent from merging into open pickings) - that might hit some bugs. I wonder if stock subsystem is really a good place to get excited with OUs instead of locations/warehouses..
- MRP and the rest of the crew: I don't see where splitting can have a use case. The only TX is `mrp.production` - Splitting a production in two? - Rather prevent merging

Note: I see how "prevent merging" on models that do have those merging functionality can become an issue. But then, we also can choose NOT to support certain transactions through OUs if it doesn't make sense (eg. on quants, which AFAIK anyway adopt the company of the location, they're currently in)

blaggacao:
Speaking of the devil: https://github.com/odoo/odoo/commit/a5b6f31cf28e5381e1c85f66730bcdb55998e643

Specifically:
```
2 property methods have been added on the environment to retrieve the company
on which the user is logged in and the companies the user enabled, on a specific
tab.
That way, when creating a record, instead of doing
default=lambda self: self.env.user.company_id
do
default=lambda self: self.env.company_id
On the other hand, to retrieve the enabled companies, do
companies = self.env.company_ids
```

blaggacao:
@gdgellatly I now might agree with you: The best approach would be to lobby Odoo into decoupling company and CoA a little. They are too "double use". That might be the root of all headaches...

blaggacao:
screenshot-518884-master-252c10 runbot16 odoo com-2019 05 13-12-25-05
The new context switcher. Nice UI :smile:

PR #194, opened 2019-05-13
[12.0] [ADD] hr_operating_unit

meigallodixital:
This module introduces Operating Units in Employees


hr_operating_unit/models/hr_employee.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    operating_unit_ids = fields.Many2many(
+        'operating.unit', 'operating_unit_employees_rel', 'employee_id',
+        'poid', 'Operating Units',
```
nikul-serpentcs, line 11: @meigallodixital Name should be poid --> `operating_unit_id`

hr_operating_unit/models/hr_operating_unit.py
```diff
@@ -0,0 +1,12 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrOperatingUnit(models.Model):
+    _inherit = 'operating.unit'
+
+    employee_ids = fields.Many2many(
+        'hr.employee', 'operating_unit_employees_rel', 'poid', 'employee_id',
```
nikul-serpentcs, line 10: same as here

nikul-serpentcs (commented):
nitpicking

hr_operating_unit/models/hr_employee.py
meigallodixital: I have called it **poid** to maintain the nomeclature of the operating unit base module [1]. I have no problem to change it.

[1] https://github.com/OCA/operating-unit/blob/91cfccae6101e9d4e90825a9d6484c9c25ada786/operating_unit/models/res_users.py#L28

AaronHForgeFlow: Yes, sorry about that, will be fixed in v13. Now it is not convenient to change that. There is no benefit and there is a migration script needed. Not worth it now.

AaronHForgeFlow:
btw if there is a related user to the employee, the OU should be the same, shouldn't they?

hr_operating_unit/models/hr_employee.py
meigallodixital: Do we confirm then that for 12.0 we keep 'poid' then?

We have also migrated hr_holidays and  ht_attendance. If we close this PR we send the other two. Same github repository on its branches.

As this PR is too old (commits are from May) could it be good to rebase and squash cmmits and send it again?

AaronHForgeFlow: Yes, keep poid in this version. Will be fixed in v13 with the proper migration script.

Regarding the other question. PRs will be accepted if they solve a generic use case that is not covered by the existing modules and the PR makes sense. For example, for holidays, the operating unit must be related to the employee. We need to make sure we don't create inconsistencies between the different models. But, of course, any contribution is welcome. :)

meigallodixital:
> 
> 
> btw if there is a related user to the employee, the OU should be the same, shouldn't they?

You can have an organizational unit as a user ('central' for example) but work (attendances) in a different organizational unit or even in several organizational units (our supervisor, for example, works one-two days in each store and we have seven + central). If you think it is better for simplicity to fix it we change it, you are the ones who know most about OU :)

What I miss is a selector to change the default OU on the gui.

As I am from spain if you want to talk this things on the maillist or any other way could be good. We can paste later here the decisions, but talk on spanish this complex things is more comfortable to me :)

AaronHForgeFlow:
@meigallodixital for now I think we understand each other for now :) 

Yes, for attendances and holidays I understand that a manager may have attendance on different OU.  I think it is correct.

However, the manager is one single employee, one single user. So if for example, Admin user is assigned to operating unit b2b and b2c. Then the employee should be assigned to the same OUs, don't you think?

meigallodixital:
> 
> 
> @meigallodixital for now I think we understand each other for now :)
> 
> Yes, for attendances and holidays I understand that a manager may have attendance on different OU. I think it is correct.
> 
> However, the manager is one single employee, one single user. So if for example, Admin user is assigned to operating unit b2b and b2c. Then the employee should be assigned to the same OUs, don't you think?

We have different points of view because in our case the OUs are physical stores. In yours it is to mark business model according to the example you say.

I reiterate the case of the supervisor. As  employee his OU does not change, it belongs to 'central' and to the main company (we have multi-company, 3 stores + 3 stores  + 1 store). But as user he must be able to make changes to any OU, and you have to change each time you make an operation the OU in which you are currently is not efficient. Therefore, I may be interested in that as an employee I am marked 'central', as a user I am marked 'store 1' and they do not match.

Just because of this I commented on the drop-down to change from default OU equal to the company one  for multi OU accounts. Even it would be interesting if it were temporary, when logging out, recover the employee default (the fixed one) into user for example.

In any case, I repeat that the ones that you know the most about how these changes can affect the global OU support are you. We have to look for the most generic change to serve everyone. If you consider changing it we do it without problems. :)

AaronHForgeFlow:
But if the supervisor belongs to central + all 3 stores, he will be able to see everything as it is now, and that is correct right?

By the way, everyone following this repo can give his opinion here :) Hope some others give their point of view.

meigallodixital:
> 
> 
> But if the supervisor belongs to central + all 3 stores, he will be able to see everything as it is now, and that is correct right?
> 
> By the way, everyone following this repo can give his opinion here :) Hope some others give their point of view.

The problem is not seeing all the data. If your default OU is not the same as the OU you are working on today, you have to select again and again the correct one in each operation. We talk about dozens of daily operations. Do not think that because he is a supervisor he does not work ;)

If I have to send an employee as reinforcement or replacement from one store to another I have the same problem. As employee your OU does not change, just that day that person works in another store, needs change user OU not employee OU. 

In physical store this is operationally unfeasible. Apart from that if you hurry up and make a mistake, buy / sell for another store (we buy and sell in the store itself, we are second hand). We need all operations tagged by OU because the stores are their own entities, such as micro businesses (franchises).

That said, I am convinced of the option to match them as you said **but** have a drop-down to change it and operate with is OU temporarily ... That will be a good and versatile solution imho. As now the only way to change the default user operating unit is by a admin if i am correct.

meigallodixital:
As it was decided not to squash the commits into one, and move them to the end of the branch with actual date, I restored the branch to leave it as it was originally. As I usually do this in my branches (rebase -i and squash the commits to only one), I ask what would be right in OCA.

What would be the best process in this case to update the forked branch? Merge when updating the repository (which leaves a merge confirmation in the history) or rebase without squash the commits?

AaronHForgeFlow:
Better to rebase, merge commits are not desirable in OCA.

Regarding your comment before, I think it would be awesome to have an easy default OU switcher on the top bar, similar to company switcher in Odoo13 or similar to https://github.com/OCA/operating-unit/pull/193#issuecomment-491912952

meigallodixital:
> 
> 
> Better to rebase, merge commits are not desirable in OCA.
> 
> Regarding your comment before, I think it would be awesome to have an easy default OU switcher on the top bar, similar to company switcher in Odoo13 or similar to [#193 (comment)](https://github.com/OCA/operating-unit/pull/193#issuecomment-491912952)

I am trying to do it. But i have a problem to update session_info. If you revise google spanish list you will see the question ...

max3903 (approved):

osi-scampbell (approved):

AaronHForgeFlow:
@meigallodixital  could you squash before merging this?

meigallodixital:
> 
> 
> @meigallodixital could you squash before merging this?

Unfortunately I have been busy at home last days. I will take a look ASAP.

patrickrwilson (changes requested):
I think the usage of 'default_operating_unit_id' and 'operating_unit_ids' might be inconsistent with the other OU modules. 
There are several simple approaches...
1. If the idea is to have a single OU on the employee records, then change 'default_operating_unit_id' to 'operating_unit_id' and change record rule to (operating_unit_id = false OR operating_unit_ids in user.operating_unit_ids). Remove operating_unit_ids from the employee entity since it's not needed.
2. If the idea is to have multiple OU's on the employee record then do the opposite of above.  Keep 'operating_unit_ids', remove 'default_operating_id'. The record rule would then be (operating_unit_ids = false OR operating_unit_ids in user.operating_unit_ids.

Either way is consistent of the other modules that I've seen in this repo.

hr_operating_unit/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module introduces the following features:
+
+* Adds the Operating Unit (OU) to the Employee.
+
```
max3903, line 4:
```suggestion
```

hr_operating_unit/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
...
+* The Employees's default Operating Unit (OU) is proposed at the time of creation
+  getting it from the logged user.
+
```
max3903, line 7:
```suggestion
```

hr_operating_unit/security/security.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903, line 1:
```suggestion
```

hr_operating_unit/views/hr_employee.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
```
max3903, line 1:
```suggestion
```

hr_operating_unit/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Gonzalo González Domínguez <meigallo@meigallodixital.com>
+
```
max3903:
```suggestion
```

max3903:
@meigallodixital Can you provide any time-frame when you will fix the comments so we can get this merged? Or should we take your branch from here and make a new pull request?

meigallodixital:
> 
> 
> @meigallodixital Can you provide any time-frame when you will fix the comments so we can get this merged? Or should we take your branch from here and make a new pull request?

I will try to do it today, sorry for the delay. These are complex days :(

meigallodixital:
> 
> 
> I think the usage of 'default_operating_unit_id' and 'operating_unit_ids' might be inconsistent with the other OU modules.
> There are several simple approaches...
> 
>     1. If the idea is to have a single OU on the employee records, then change 'default_operating_unit_id' to 'operating_unit_id' and change record rule to (operating_unit_id = false OR operating_unit_ids in user.operating_unit_ids). Remove operating_unit_ids from the employee entity since it's not needed.
> 
>     2. If the idea is to have multiple OU's on the employee record then do the opposite of above.  Keep 'operating_unit_ids', remove 'default_operating_id'. The record rule would then be (operating_unit_ids = false OR operating_unit_ids in user.operating_unit_ids.
> 
> 
> Either way is consistent of the other modules that I've seen in this repo.

meigallodixital:
> 
> 
> > I think the usage of 'default_operating_unit_id' and 'operating_unit_ids' might be inconsistent with the other OU modules.
> > There are several simple approaches...
> > ```
> > 1. If the idea is to have a single OU on the employee records, then change 'default_operating_unit_id' to 'operating_unit_id' and change record rule to (operating_unit_id = false OR operating_unit_ids in user.operating_unit_ids). Remove operating_unit_ids from the employee entity since it's not needed.
> > 
> > 2. If the idea is to have multiple OU's on the employee record then do the opposite of above.  Keep 'operating_unit_ids', remove 'default_operating_id'. The record rule would then be (operating_unit_ids = false OR operating_unit_ids in user.operating_unit_ids.
> > ```
> > 
> > 
> > Either way is consistent of the other modules that I've seen in this repo.

Sorry, I accidentally closed it, I don't see how to reopen it.

The reaseon why both exist was discussed with Aaron above. In short, an employee can be in multiple authorized OUs but currently in one. For example, in the case that your supervisor is the one assigned to you, the one you have to cover among the ones you have. It is also necessary to avoid having to cover the OU in which you are working in each operation. It is explained more extensively as I say more messages above.

In case it is decided that one of them should disappear, it should be ids. Default I see it is necessary, so as not to have to repeat in each operation that ou is in employee on.

max3903:
Anything blocking this one from being merged?

meigallodixital:
> 
> 
> Anything blocking this one from being merged?

The previous debate on whether to leave only one of the fields or both is pending. We see it as necessary, since one parameter is the default OU and the other is which OUs you can operate on (it doesn't always have to be one). But the community decides :)

max3903:
+1 to have 2 fields: (default OU + allowed OUs)

dreispt:
Reviewers: I' will merge this soon unless I get a vote against in the next days.
Thank you.

PR #195, opened 2019-05-13
[12.0] [ADD] hr_attendance_operating_unit

meigallodixital:
This module introduces Operating Units in Attendances

On kiosk mode, the operating unit of the attendance is selected from the employee because it is executed by __system__ user. If the operating units are used to know in which company's area the employee makes the attendance is not valid. We think it deserves to discuss this behavior.


AaronHForgeFlow:
HI @meigallodixital Is this suppose to track the time spent in one or another operating unit? If it is that, maybe we can just use  the timesheets for that. Because the analytic account has the operating unit assigned.

meigallodixital:
> 
> 
> HI @meigallodixital Is this suppose to track the time spent in one or another operating unit? If it is that, maybe we can just use the timesheets for that. Because the analytic account has the operating unit assigned.

No. On Kiosk Mode you see all company employee's to select when you make an attendance. This module is for limit that view to only employees (attendances) on the user logged OU(s). And to group the assistances by OU. This is only for attendances have the employee worksheets or not.

azvolamza (approved):

AaronHForgeFlow:
Ok I understand. I am wondering if it is possible in the rule to check if the operating unit of the user matches the operating unit of the attendance. That way perhaps we don't need to change the data model.

<field name="domain_force">[('employee_id.user_id.operating_unit_ids','in', user.operating_unit_ids.ids)]</field>

azvolamza:
how>>>>>

AaronHForgeFlow:
I mean a ir.rule in hr_attendance with a domain like:
`<field name="domain_force">[('employee_id.user_id.operating_unit_ids','in', user.operating_unit_ids.ids)]/>`

Did not try that thought.

meigallodixital:
> 
> 
> Ok I understand. I am wondering if it is possible in the rule to check if the operating unit of the user matches the operating unit of the attendance. That way perhaps we don't need to change the data model.


> I mean a ir.rule in hr_attendance with a domain like:




> 
> 
> Ok I understand. I am wondering if it is possible in the rule to check if the operating unit of the user matches the operating unit of the attendance. That way perhaps we don't need to change the data model.
> 
> [('employee_id.user_id.operating_unit_ids','in', user.operating_unit_ids.ids)]

The problem is that in the kiosk mode the user in session  is **\_\_system\_\_** (odoobot) not the real user. That overwriting of the create method is a workaround for that case. If the user is **\_\_system\_\_**, we will get the employee's default value instead.

I don't know if there are any way to get **real** user logged in that case.

PR #223, opened 2019-12-29
[13.0][MIG] operating unit

alan196:


operating_unit/models/operating_unit.py
```diff
@@ -0,0 +1,63 @@
...
+    active = fields.Boolean(default=True)
+    company_id = fields.Many2one(
+        "res.company",
+        required=True,
+        readonly=True,
+        default=lambda self: self.env.company,
+    )
+    partner_id = fields.Many2one("res.partner", "Partner", required=True)
+    user_ids = fields.Many2many(
+        "res.users", "operating_unit_users_rel", "poid", "user_id", "Users Allowed"
```
AaronHForgeFlow: I think version 13 is a good moment to put a good column name here. It should be something like `operating_unit_id` or `ou_id`. Migration script could be added later, after the merge.

AaronHForgeFlow (commented):
@alan196 Thanks for migrating this. It seems to work fine. Could you split the migration commit in two? First commit with the isort,black and all the syntax stuff. And the second one with the actual migration. That is easy for reviewing.

Thanks.

operating_unit/models/operating_unit.py
alan196: Thanks for your review! I already rename the column to operating_unit_id

alan196:
@AaronHForgeFlow I already split the commit in two. 👍

operating_unit/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright 2015-TODAY Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
+# License: LGPL-3 or later (https://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Operating Unit",
+    "summary": "An operating unit (OU) is an organizational entity part of a "
+    "company",
+    "version": "13.0.1.0.0",
+    "author": "Eficent, "
+    "Serpent Consulting Services Pvt. Ltd.,"
+    "Jarsa Sistemas, "
```
AaronHForgeFlow: For migrating modules without adding any major feature putting yourself in the contributors section in the readme should be enough. This is reserved for the original author + contributors who do major functional changes.

Hope you understand.

alan196: Thanks for the clarification 👍
I already make the change

AaronHForgeFlow: Thank you!

AaronHForgeFlow (approved):
:+1: Tested + code. Thanks.

LoisRForgeFlow (approved):
LGTM :+1:

lmgr0312 (approved):
LGTM.. :+1:

BT-jortells:
@alan196 @AaronHForgeFlow @LoisRForgeFlow 

The issues specified in [V12 Known issues / Roadmap](https://github.com/OCA/operating-unit/blob/12.0/operating_unit/README.rst#known-issues--roadmap) have been solved only partially in this PR, since the migration script has not been developed. So, an update from v12 to v13 will result in an error because of the table column name changed

AaronHForgeFlow:
@BT-jortells Yes, you are right. It is still pending. I don't have time right now for this, as I am not migrating to v13 now. But, eventually, it has to be done.

BT-jortells:
Thanks for answering. I have also referenced this in the [Migration to version 13.0 issue](https://github.com/OCA/operating-unit/issues/200) to keep it in mind.

PR #225, opened 2019-12-30
[12.0] [ADD] product_operating_unit

nikul-serpentcs:
Product Operating Unit
====================

- It introduces the operating unit to the product template.
- The operating unit from the user is assigned by default when creating a new
  product template.
- In case of multi-company, no operating unit from another company to that
  assigned to the product can be set for it
- It implements user's security rules.


max3903:
@Dangar-Serpentcs We also need the `operating_unit_ids` field on `product.category`. The field on `product.template` should inherit the OUs from the category, and may have additional OUs.

We need to:
- [x] add a method `onchange_product_category` to `product.template` to add the OUs of the category to the product
- [x] add a constraint method `check_operating_unit` to make sure that the  OUs of the category are on the product

This should allow @BT-rmartin to configure specific OUs on products.

nikul-serpentcs:
@max3903 FYI  here Travis Failed, because of `stock_account_operating_unit` used Product and in product category has no OU's.

we are using this Case
> add a constraint method check_operating_unit to make sure that the OUs of the category are on the product

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+    @api.constrains('operating_unit_ids', 'company_id')
+    def _check_company_operating_unit(self):
+        for record in self:
+            if record.company_id and record.operating_unit_ids:
+                for operating_unit in record.operating_unit_ids:
+                    if record.company_id != operating_unit.company_id:
+                        raise ValidationError(
+                            _('Configuration error. The Company in the Product'
+                              ' and in the Operating Unit must be the same.')
+                        )
```
max3903: @nikul-serpentcs This constraint is not necessary. Please remove it.

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @api.multi
+    @api.constrains('operating_unit_ids', 'categ_id')
+    def _check_operating_unit(self):
+        for record in self:
+            if record.operating_unit_ids and record.operating_unit_ids != \
+                    record.categ_id.operating_unit_ids:
+                raise ValidationError(_(
+                    '''The Operating Unit in the Product and in the
+                    Category must be the same.'''))
```
max3903: @nikul-serpentcs The OU don't need to be the same. The OU of the product needs to include the ones from the category.

Category 1: OU 1, OU 2
Product in Category 1: OU 1, OU 2, OU 3

max3903:
@nikul-serpentcs Please review and merge https://github.com/nikul-serpentcs/operating-unit/pull/1

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,41 @@
...
+                    from the category.'''))
+
+    @api.multi
+    @api.onchange('operating_unit_ids', 'categ_id')
+    def onchange_operating_unit_ids(self):
+        for record in self:
+            if record.categ_id:
+                record.categ_id.write({
+                    'operating_unit_ids':
+                        [(6, 0, record.operating_unit_ids.ids)]})
```
Dangar-Serpentcs: Don't **write** record in **onchange** method because if user change the OU and after that **discard** the record it will set the value of OU without saving record.

max3903:
@nikul-serpentcs Please review and merge https://github.com/nikul-serpentcs/operating-unit/pull/2

product_operating_unit/models/product_template.py
max3903: Fixed in https://github.com/nikul-serpentcs/operating-unit/pull/2

murtuzasaleh:
> @max3903 FYI here Travis Failed, because of `stock_account_operating_unit` used Product and in product category has no OU's.
> 
> we are using this Case
> 
> > add a constraint method check_operating_unit to make sure that the OUs of the category are on the product

@max3903

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,40 @@
...
+                    "The operating units of the product must include the "
+                    "ones from the category."))
+
+    @api.multi
+    @api.onchange('categ_id')
+    def onchange_operating_unit_ids(self):
+        for record in self:
+            if record.categ_id:
+                record.operating_unit_ids = \
+                    [(6, 0, record.operating_unit_ids.ids)]
```
Dangar-Serpentcs: Hey @max3903 do you think this **Onchange** work well here, I don't think so.
I think you need to improve once again to set Operating units to the Product Template from product category
**ie.**

    @api.multi
    @api.onchange('categ_id')
    def onchange_operating_unit_ids(self):
        for record in self:
            if record.categ_id:
                record.operating_unit_ids = \
                    [(6, 0, record.categ_id.operating_unit_ids.ids)]

max3903: @nikul-serpentcs Please review and merge https://github.com/nikul-serpentcs/operating-unit/pull/3

product_operating_unit/models/product_category.py
```diff
@@ -0,0 +1,24 @@
...
+    operating_unit_ids = fields.Many2many(
+        'operating.unit',
+        'product_category_operating_unit_rel',
+        string='Operating Units')
+
+    @api.multi
+    def write(self, vals):
+        if vals.get('operating_unit_ids'):
+            products = self.env['product.template'].search([
+                ('categ_id', 'child_of', self.id)])
```
BT-rmartin: self could be a recordset with many categories, then this approach is wrong. You should iterate on self first.

product_operating_unit/models/product_category.py
```diff
@@ -0,0 +1,24 @@
...
+            for product in products:
+                ou_ids = product.operating_unit_ids.ids
+                ou_ids.extend(vals.get('operating_unit_ids')[0][2])
+                product.operating_unit_ids = [(6, 0, ou_ids)]
+        return super().write(vals)
```
BT-rmartin: Why not super(ProductCategory, self).write(vals)?
Why not doing that before the code above (linesn 17-23) so that if it fails we are not changing the products?

product_operating_unit/models/product_template.py
```diff
@@ -18,18 +15,25 @@ def _default_operating_unit_ids(self):
...
+    def onchange_operating_unit_ids(self):
         for record in self:
-            if record.company_id and record.operating_unit_ids:
-                for operating_unit in record.operating_unit_ids:
-                    if record.company_id != operating_unit.company_id:
-                        raise ValidationError(
-                            _('Configuration error. The Company in the Product'
-                              ' and in the Operating Unit must be the same.')
-                        )
+            record.operating_unit_ids = \
```
BT-rmartin, line 42: When changing the category you are resetting the ous to those coming from there but you are removing the possible other ones added for the product that differ from those of the category. Could it be changed so that it leaves those which weren't in the old category taking care of not duplicating them if they are in the new category?

product_operating_unit/models/product_template.py
```diff
@@ -18,18 +15,25 @@ def _default_operating_unit_ids(self):
...
 
     operating_unit_ids = fields.Many2many(
-        'operating.unit', 'product_operating_unit_rel',
-        string='Operating Unit',
-        default=_default_operating_unit_ids
-    )
+        'operating.unit',
+        'product_operating_unit_rel',
+        string='Operating Units',
+        default=_default_operating_unit_ids)
```
BT-rmartin, line 25: With this new approach, the default ones shouldn't be the ones from the category instead of those from the user?
Apart from that, for existing products, I guess we need a migration script that initializes them.

product_operating_unit/security/product_template_security.xml
```diff
@@ -1,12 +1,15 @@
 <!--
     Copyright (c) 2019 brain-tec AG (http://www.braintec-group.com)
+    Copyright (C) 2019 Open Source Integrators
+    Copyright (C) 2019 Serpent Consulting Services
     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 -->
 <odoo>
 
     <record id="ir_rule_product_template_allowed_operating_units" model="ir.rule">
```
BT-rmartin, line 9: Now we will probably need a new rule for product category.

product_operating_unit/tests/test_product_operating_unit.py
```diff
@@ -39,8 +37,12 @@ def setUp(self):
                                           [self.group_stock_user],
                                           self.company2,
                                           [self.b2b])
+        self.product1.categ_id.operating_unit_ids = [(6, 0, [self.ou1.id])]
         self.product1.operating_unit_ids = [(6, 0, [self.ou1.id])]
```
BT-rmartin: If doing the changes I proposed before, this line is not needed, as the ous in the product should come from the category. Same comment for line 43

BT-rmartin (changes requested):
In general, I do think this is a good input, so thanks for your work. 
Nevertheless, please, have a look at my comments. Some changes are needed from my point of view. 
We can discuss about them if needed.

nikul-serpentcs:
@BT-rmartin Code updated as per your comment, Can you please review it?

BT-rmartin:
@nikul-serpentcs Please, small change to be done. All the other things are ok to me
Also a pending comment from previous review. Please, add you opinion on it if you won't address it

product_operating_unit/models/product_template.py
nikul-serpentcs: Yes, I agree @BT-rmartin  
When I removing OU in a category from one product but those categories are defined in other products than what should I Do?

Case 1:
category_1 have OU1 And OU2
Product_1 have category_1, so Product 1 has OU1 and OU2

Now, I removed category_1 in OU2
In product_2 have category_1, so Product 2 has only OU1

That Time Product_1 has OU1 and OU2, it's Right?

CC @max3903

BT-rmartin (approved):

product_operating_unit/models/product_template.py
BT-rmartin: Your example is another use case. 
When you created Product1 the category_1 had OU1 and OU2. 
This means you weren't adding any OU in addition, but you just took the ones the category had. 
Now you are suggesting a new use case when you remove OU2 from category.

In this case you should update all products accordingly, not only new ones, but also existing ones. 
The reason is that even if you added it manually on the product when it wasn't in the category, in the end, you need to have added it to the category (and consequently to the product) to be able to remove it from the category (and consequently to the product).
 
Imagine the next scenario

Category_1: OU1 and OU2
Create Product1: It will have OU1 and OU2 from Category_1. Now imagine we add manually OU3
Edit Category_1: Remove OU2. Consequently you automatically removes from product OU2, which will keep OU1 and OU3
Edit Product1: You add OU2 manually, then it has again OU1, OU2 and OU3
Edit Category_1: You add OU2. Consequently it tries to add OU2 to the product, but it realizes it is already there, so no action
Edit Category_2: You remove OU2. Consequently you automatically remove from product OU2, but this time, it was added manually by user. Anyway it is removed as it might have been also added by previous change in Category_1, so the product will have OU1 and OU3

Then I would say that's ok. We can add and remove according to category operating units taken care of not creating duplicates and respecting the OUs added by the user not existing at any time in the category.

BT-rmartin (changes requested):
Please, have a look at my comments for the open point. Thanks

max3903:
@BT-rmartin 

When adding an OU to a category, it is clear that you want to add the OU to all the products in that category.

When removing, it is another story because some users in that OU may be using the product in their transactions. I think it is safer in this case to leave the products as is and leave it to the user to remove the OU on the products manually.

max3903:
@nikul-serpentcs Can you fix the Travis errors please?

nikul-serpentcs:
> @nikul-serpentcs Can you fix the Travis errors please?

@max3903 Travis Is Fixed, But Some doubt in Product Templates  `ir.rule`.

Here we Improving Product Templates Rule because of another domain was a clash or value was wrong.

When I comment code https://github.com/OCA/operating-unit/pull/225/commits/19f7a343c3afb5ed9b6ae8e99dcbd0fe2c2a4fda#diff-755bb6949c8d2bf977be42feb247364cR9 Travis is Green :green_apple:

max3903:
@nikul-serpentcs What is the error if we uncomment those lines?

nikul-serpentcs:
> @nikul-serpentcs What is the error if we uncomment those lines?

@max3903 Please Check https://travis-ci.org/OCA/operating-unit/jobs/639318497#L2578
When I Uncomment Rule Got Error Like this.

osi-scampbell:
@nikul-serpentcs I think we have a dependency error here, currently this module depends on "product", and "stock". We need this to depend on operating_unit or else we cannot create those relationships to operating_unit_ids

product_operating_unit/views/product_template_view.xml
```diff
@@ -0,0 +1,48 @@
...
+<odoo>
+
+    <record id="product_template_tree_view" model="ir.ui.view">
+        <field name="name">product.template_tree</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_tree_view" />
+        <field name="arch" type="xml">
+            <field name="categ_id" position="after">
+                <field name="operating_unit_ids"
+                       domain="[('user_ids', 'in', uid)]"
```
patrickrwilson, line 17:
```suggestion
                       widget="many2many_tags"
                       domain="[('user_ids', 'in', uid)]"
```

patrickrwilson (changes requested):

BT-rmartin:
@max3903 I could agree that removing an OU from category shouldn't imply changes in the product, as it's risky. 
But when changing the category of a product, in this case, I would say it's needed to remove the OUs from the previous category and add the ones of the new category, or perhaps it's the same case as before and if product was already used this can lead to troubles. 
Then we just focus on adding on any case: either when changing category of the product or when adding OUs to the category. 
@nikul-serpentcs I don't get why you need to remove that rule. We need a rule for product template, and I don't see why that existing domain is wrong. But about the error of travis with the rule is due to some failing test that should be debugged to understand the reason behind

HirenDangar:
@nikul-serpentcs Are you going to continue on this PR ?

nikul-serpentcs:
@BT-rmartin Please check Comments

https://github.com/OCA/operating-unit/pull/225#issuecomment-575122513

Please check Error here: 
https://github.com/OCA/operating-unit/pull/225#issuecomment-576101065

max3903:
@nikul-serpentcs Looks like you need to set the OUs on products using a hook.

See an example here:
https://github.com/OCA/operating-unit/pull/274/files

murtuzasaleh (approved):

max3903 (approved):

PR #228, opened 2019-12-30
[13.0][MIG] analytic operating unit

alan196:
It depends on this PR #223

PR #229, opened 2019-12-31
[13.0][MIG] Migration of module account_operating_unit

alan196:
This PR depends on #228


alan196:
I make the migration of this module in this PR.

This module now installs and it seems to work as expected. I only have a question. If someone could help me to answer.

- I see that in the `post` method of `account.move` it has the following code.

```python
    def post(self):
        ml_obj = self.env["account.move.line"]
        for move in self:
            if not move.company_id.ou_is_self_balanced:
                continue

            # If all move lines point to the same operating unit, there's no
            # need to create a balancing move line
            ou_list_ids = [
                line.operating_unit_id and line.operating_unit_id.id
                for line in move.line_ids
                if line.operating_unit_id
            ]
            if len(ou_list_ids) <= 1:
                continue
```

But it also has the following constraint.

```python
    @api.constrains("operating_unit_id", "move_id")
    def _check_move_operating_unit(self):
        for rec in self:
            if (
                rec.move_id
                and rec.move_id.operating_unit_id
                and rec.operating_unit_id
                and rec.move_id.operating_unit_id != rec.operating_unit_id
            ):
                raise UserError(
                    _(
                        "Configuration error. The Operating Unit in"
                        " the Move Line and in the Move must be the"
                        " same."
                    )
                )
```

So there is no way you can get two `account.move.line` with different `operating_unit_id` in the same account move.
The self-balance move line will never be created.

I'm doing something wrong?

AaronHForgeFlow:
@alan196 For automatic account moves (almost all) the operating unit will be only in the account move lines.  

The only case where you will find the operating unit  in the account move is when the user creates a manual journal entry. And this is only for the default value when adding lines to the journal entry. So the user puts default  operating unit in the account move and then all the account move lines will be defaulted to that operating unit.

If the manual entry contains account move lines for different operating units then the operating unit in the account move must be blank.

BT-nstuder:
Hi @alan196, is the PR still `WIP`? If yes, are you going to update it?
Regards

AaronHForgeFlow:
@BT-nstuder  It seems it is WIP, there are still many references to the deprecated `account.invoice` model.

alan196:
@BT-nstuder  
The migration functionality works (It may need some testing).
The only thing that is pending are the tests, that's the reason you see references of account.invoice

I had a charge of work that doesn't allow me to continue with this migration until next week (February 5 to be more precise).

I will continue with this migration on that date.
If you are interested I can give you access to write in that repo and make updates.

alan196:
@AaronHForgeFlow @BT-nstuder I have adapted the tests and it is ready to be reviewed.

Could you help me to review this PR?

Thanks in advance

AaronHForgeFlow (changes requested):
I configured the company to have balanced operating units and I created an invoice for Main Operating unit.

Then I paid the invoice using a Bank for a different operating unit. However, the inter-OU accounting entries were not created.

alan196:
> I configured the company to have balanced operating units and I created an invoice for Main Operating unit.
> 
> Then I paid the invoice using a Bank for a different operating unit. However, the inter-OU accounting entries were not created.

@AaronHForgeFlow I make tests to validate this functionality 👍 now that is fixed and with tests.

I added more tests to the payment.

- [x]  Pay two invoices with payment wizard, odoo make a payment per invoice.
- [x] Make a transfer payment between banks in different OU
- [ ] Pay two invoices with payment wizard and group payment.

The latest test is created but the logic to comply with it is still missing, at this moment I'm stuck.
When a payment is done with several invoices of the same OU is not a problem.
But the problems that I define in the test is one invoice from OU1 another from OU2 paid from a bank of OU3.
I imagine two options:
- Split the account move line of the payment for each invoice with their corresponding operating unit and invoice amount. (this must consider multicurrency).
- Raise an error and don't allow the payment of invoices from different OU (I don't think it is a good option).

There is a similar case for invoices from different OU, if you will create a write-off to that payment, what OU needs to be defined in that line???

alan196:
@ougc27

ps-tubtim:
@alan196 Please check Travis.

AaronHForgeFlow:
Sorry @alan196  did not see your comment above.

> I imagine two options:
Split the account move line of the payment for each invoice with their corresponding operating unit and invoice amount. (this must consider multicurrency).
Raise an error and don't allow the payment of invoices from different OU (I don't think it is a good option).
There is a similar case for invoices from different OU, if you will create a write-off to that payment, what OU needs to be defined in that line???

IMO I think option 1 (split the payment) is the correct one. Nevertheless this options is not implemented in previous version, so in order to move forward, maybe we can include this in the roadmap, and for now, just raise an error and don'nt let create single payments for several different OUs.

Regarding the write-off OU, I think it should go to the OU of the invoice not the payment, but I need to investigate more. 

Thanks for the work!

alan196:
> Sorry @alan196 did not see your comment above.
> 
> > I imagine two options:
> > Split the account move line of the payment for each invoice with their corresponding operating unit and invoice amount. (this must consider multicurrency).
> > Raise an error and don't allow the payment of invoices from different OU (I don't think it is a good option).
> > There is a similar case for invoices from different OU, if you will create a write-off to that payment, what OU needs to be defined in that line???
> 
> IMO I think option 1 (split the payment) is the correct one. Nevertheless this options is not implemented in previous version, so in order to move forward, maybe we can include this in the roadmap, and for now, just raise an error and don'nt let create single payments for several different OUs.
> 
> Regarding the write-off OU, I think it should go to the OU of the invoice not the payment, but I need to investigate more.
> 
> Thanks for the work!

@AaronHForgeFlow 
I agree to continue with the current workflow, but I'm against a raise, what do you think if I add a system parameter, to allow or not the payment with different OU. And document it in Installation instructions.

account_operating_unit/README.rst
```diff
@@ -0,0 +1,141 @@
+===============================
```
newtratip, line 1: Could you generate readme to 13.0 ?

account_operating_unit/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools.translate import _
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    inter_ou_clearing_account_id = fields.Many2one(
+        "account.account",
```
newtratip:
```suggestion
        comodel_name="account.account",
```

account_operating_unit/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+        "Inter-operating unit\
+                                                   clearing account",
```
newtratip: string="Inter-operating unit clearing account"

account_operating_unit/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
```
newtratip: Remove empty line.

account_operating_unit/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+    ou_is_self_balanced = fields.Boolean(
+        "Operating Units are self-balanced",
```
newtratip:
```suggestion
        string="Operating Units are self-balanced",
```

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("move_id", False):
+                move = self.env["account.move"].browse(vals["move_id"])
+                if move.operating_unit_id:
+                    vals["operating_unit_id"] = move.operating_unit_id.id
+        return super(AccountMoveLine, self).create(vals_list)
```
ps-tubtim:
```suggestion
        return super().create(vals_list)
```

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+
+    @api.model
+    def _query_get(self, domain=None):
+        if domain is None:
+            domain = []
+        if self._context.get("operating_unit_ids", False):
+            domain.append(
+                ("operating_unit_id", "in", self._context.get("operating_unit_ids"))
+            )
+        return super(AccountMoveLine, self)._query_get(domain)
```
ps-tubtim:
```suggestion
        return super()._query_get(domain)
```

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+                and rec.operating_unit_id
+                and rec.move_id.operating_unit_id != rec.operating_unit_id
+            ):
+                raise UserError(
+                    _(
+                        "Configuration error. The Operating Unit in"
+                        " the Move Line and in the Move must be the"
+                        " same."
+                    )
+                )
```
ps-tubtim, line 67: IMO, I think you should check all move types. Why you check `rec.move_id.type == "entry"`?

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+        if (
+            self._context.get("default_type", False)
+            and self._context.get("default_type") != "entry"
+        ):
+            return self.env["res.users"].operating_unit_default_get()
+        return False
+
+    operating_unit_id = fields.Many2one(
+        comodel_name="operating.unit",
+        default=_default_operating_unit_id,
```
ps-tubtim, line 84: You can use `default=lambda self: self.env["res.users"].operating_unit_default_get(),` instead this code.

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+        states={"draft": [("readonly", False)]},
+    )
+
+    @api.onchange("invoice_line_ids")
+    def _onchange_invoice_line_ids(self):
+        res = super(AccountMove, self)._onchange_invoice_line_ids()
+        for line in self.line_ids:
+            if self.operating_unit_id:
+                line.operating_unit_id = self.operating_unit_id
+        return res
```
ps-tubtim:
```suggestion
    @api.onchange("invoice_line_ids")
    def _onchange_invoice_line_ids(self):
        if self.operating_unit_id:
            for line in self.line_ids:            
                line.operating_unit_id = self.operating_unit_id
        return super()._onchange_invoice_line_ids()
```

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+                    move, ou_id, ou_balances
+                )
+                if line_data:
+                    amls.append(ml_obj.with_context(wip=True).create(line_data))
+            if amls:
+                move.with_context(wip=False).write(
+                    {"line_ids": [(4, aml.id) for aml in amls]}
+                )
+
+        return super(AccountMove, self).post()
```
ps-tubtim:
```suggestion
        return super().post()
```

account_operating_unit/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _check_balanced(self):
+        if self.env.context.get("wip"):
+            return True
+        return super(AccountMove, self)._check_balanced()
```
ps-tubtim:
```suggestion
        return super()._check_balanced()
```

account_operating_unit/models/account_payment.py
```diff
@@ -0,0 +1,42 @@
...
+
+    operating_unit_id = fields.Many2one(
+        "operating.unit",
+        domain="[('user_ids', '=', uid)]",
+        compute="_compute_operating_unit_id",
+        store=True,
+    )
+
+    def _prepare_payment_moves(self):
+        res = super(AccountPayment, self)._prepare_payment_moves()
```
ps-tubtim:
```suggestion
        res = super()._prepare_payment_moves()
```

account_operating_unit/models/res_company.py
ps-tubtim:
```suggestion
        "Inter-operating unit clearing account",
```

account_operating_unit/report/account_invoice_report.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class AccountInvoiceReport(models.Model):
+
+    _inherit = "account.invoice.report"
+
+    operating_unit_id = fields.Many2one("operating.unit", "Operating Unit")
+
+    def _select(self):
+        select_str = super(AccountInvoiceReport, self)._select()
```
ps-tubtim:
```suggestion
        select_str = super()._select()
```

account_operating_unit/report/account_invoice_report.py
```diff
@@ -0,0 +1,26 @@
...
+        select_str += """
+            ,line.operating_unit_id
+        """
+        return select_str
+
+    def _group_by(self):
+        group_by_str = super(AccountInvoiceReport, self)._group_by()
```
ps-tubtim:
```suggestion
        group_by_str = super()._group_by()
```

account_operating_unit/views/account_payment_view.xml
```diff
@@ -0,0 +1,69 @@
...
+        <field name="model">account.payment</field>
+        <field name="inherit_id" ref="account.view_account_payment_search" />
+        <field name="arch" type="xml">
+            <field name="journal_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    groups="operating_unit.group_multi_operating_unit"
+                />
+            </field>
+            <field name="journal_id" position="after">
```
newtratip:
```suggestion
            
```

account_operating_unit/views/account_payment_view.xml
```diff
@@ -0,0 +1,69 @@
...
+        <field name="name">account.payment.search</field>
+        <field name="model">account.payment</field>
+        <field name="inherit_id" ref="account.view_account_payment_search" />
+        <field name="arch" type="xml">
+            <field name="journal_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    groups="operating_unit.group_multi_operating_unit"
+                />
+            </field>
```
newtratip:
```suggestion
            
```

account_operating_unit/views/account_payment_view.xml
```diff
@@ -0,0 +1,69 @@
...
+            <field name="journal_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    options="{'no_create': True}"
+                    groups="operating_unit.group_multi_operating_unit"
+                />
+            </field>
+        </field>
+    </record>
+    <record id="view_account_supplier_payment_tree" model="ir.ui.view">
```
newtratip: Supplier Payment Tree is show 2 columns for Operating Unit. You don't need to add this record because view_account_supplier_payment_tree inherited view_account_payment_tree which have column Operating Unit already in record above.
Selection_118

newtratip (commented):
Some Review.

account_operating_unit/views/account_move_view.xml
```diff
@@ -0,0 +1,85 @@
...
+                />
+            </xpath>
+        </field>
+    </record>
+    <record id="view_account_move_line_filter" model="ir.ui.view">
+        <field name="name">Journal Items</field>
+        <field name="model">account.move.line</field>
+        <field name="inherit_id" ref="account.view_account_move_line_filter" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='account_id']" position="after">
```
newtratip:
```suggestion
            
```

account_operating_unit/views/account_move_view.xml
```diff
@@ -0,0 +1,85 @@
...
+                <filter name="account" position="after">
+                    <filter
+                        string="Operating Unit"
+                        name="operating_unit_grouped"
+                        icon="terp-folder-green"
+                        context="{'group_by':'operating_unit_id'}"
+                        groups="operating_unit.group_multi_operating_unit"
+                    />
+                </filter>
+            </xpath>
```
newtratip:
```suggestion
            
```

account_operating_unit/views/account_move_view.xml
```diff
@@ -0,0 +1,85 @@
...
+        <field name="inherit_id" ref="account.view_account_move_line_filter" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='account_id']" position="after">
+                <field name="account_id" position="after">
+                    <field
+                        name="operating_unit_id"
+                        groups="operating_unit.group_multi_operating_unit"
+                    />
+                </field>
+                <filter name="account" position="after">
```
newtratip:
```suggestion
                <filter name="group_by_account" position="after">
```

account_operating_unit/views/account_move_view.xml
```diff
@@ -0,0 +1,85 @@
...
+        <field name="arch" type="xml">
+            <field name="journal_id" position="after">
+                <field name="operating_unit_id" />
+            </field>
+            <field name="line_ids" position="attributes">
+                <attribute name="context">{
+                    'line_ids': line_ids,
+                    'journal_id': journal_id,
+                    'default_operating_unit_id': operating_unit_id}
+                </attribute>
```
newtratip, line 75: I think context should copy from original view and add `'default_operating_unit_id': operating_unit_id` at last this.
Selection_119

newtratip:
@alan196 Could you add Operating Unit in tree view and search view of invoice ?

account_operating_unit/views/account_move_view.xml
```diff
@@ -0,0 +1,85 @@
...
+            </xpath>
+        </field>
+    </record>
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">account.move.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <field name="journal_id" position="after">
+                <field name="operating_unit_id" />
```
newtratip:
```suggestion
                <field name="operating_unit_id" options="{'no_create': True}" groups="operating_unit.group_multi_operating_unit" />
```

account_operating_unit/security/account_security.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
newtratip, line 1: Add rule for account move too.

account_operating_unit/report/account_invoice_report.py
```diff
@@ -0,0 +1,26 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class AccountInvoiceReport(models.Model):
+
+    _inherit = "account.invoice.report"
+
+    operating_unit_id = fields.Many2one("operating.unit", "Operating Unit")
```
newtratip:
```suggestion
    operating_unit_id = fields.Many2one(comodel_name="operating.unit", string="Operating Unit")
```

account_operating_unit/models/account_payment.py
```diff
@@ -0,0 +1,42 @@
...
+    _inherit = "account.payment"
+
+    @api.depends("journal_id")
+    def _compute_operating_unit_id(self):
+        for payment in self:
+            if payment.journal_id:
+                payment.operating_unit_id = payment.journal_id.operating_unit_id
+
+    operating_unit_id = fields.Many2one(
+        "operating.unit",
```
newtratip:
```suggestion
        comodel_name="operating.unit",
```

account_operating_unit/models/res_company.py
```diff
@@ -0,0 +1,36 @@
...
+        help="Activate if your company is "
+        "required to generate a balanced"
+        " balance sheet for each "
+        "operating unit.",
+    )
+
+    @api.constrains("ou_is_self_balanced")
```
newtratip:
```suggestion
    @api.constrains("ou_is_self_balanced", "inter_ou_clearing_account_id")
```
IMO, If you change inter_ou_clearing_account_id to empty and ou_is_self_balanced still check. The system should show raise error too.

AaronHForgeFlow (commented):
The inter operating unit account entries are created correctly when paying invoices from different OU. For me it is ok to move forward like it is now. Just pending the CI checks to proceed to merge.

AaronHForgeFlow (approved):
:+1:  Functional tested. Thank you!

lmgr0312 (approved):
LGTM!

AaronHForgeFlow:
Will merge when comments are attended.

account_operating_unit/models/account_move.py
alan196: This test is only valid when you add the account move manually, as commented by @AaronHForgeFlow in this comment https://github.com/OCA/operating-unit/pull/229#issuecomment-570145352 that's the reason this constraint only validate moves when are of type `entry`

account_operating_unit/models/account_move.py
alan196: This is related with the comment https://github.com/OCA/operating-unit/pull/229/files#r408973265

You don't need a default operating unit to be defined when the type is entry

AaronHForgeFlow:
All comments attended. Thank you @alan196 !

PR #233, opened 2020-01-13
[12.0][FIX] stock_security

AaronHForgeFlow:
bad `ir.rule` is denying access to stock.picking.type wrongly


AdriaGForgeFlow (approved):

JoanSForgeFlow (approved):

LoisRForgeFlow:
What is the difference? I see it equivalent

AaronHForgeFlow:
> What is the difference? I see it equivalent

I think the same but they are not :fearful: I cannot access picking types with the current ir.rule, and I do with the update ir.rule.

PR #235, opened 2020-01-15
[IMP] purchase_operating_unit: PO-OU's Multi Ware Access

osi-scampbell:
This PR allows all PO's to access warehouses that do not have operating_unit_id's set. It also adjusts the default_picking_type method to avoid getting an id from a null value.

#237


AaronHForgeFlow (approved):
:+1: LGTM

max3903 (approved):

bodedra (approved):

PR #238, opened 2020-01-15
[FIX] account_operating_unit: Ensure Company Set

osi-scampbell:
When removing 'editable="bottom"' from the form view, some users have encountered a bug where company_id is not set until after the form is saved. We were able to replicate this behavior on runbot here:
image

To ensure this doesn't happen, we set the default company_id through the context.


AaronHForgeFlow (approved):
:+1: thanks

max3903 (approved):

bodedra (approved):

PR #239, opened 2020-01-17
[FIX] res_partner_operating_unit: Sync Contact/User + Vals Not Res

osi-scampbell:
In res_partner_operating_unit when you change the default_operating_unti_id you will recieve an error because super().write(vals) returns True not an existing record. So when we go to write to that record we get the following error: AttributeError: 'bool' object has no attribute 'default_operating_unit_id'

In the method below we append that default_operating_unit_id to the corresponding partner record however we do not append it to the user record itself which creates confusion. This way, changing the default OU adds to both partner and user record.


res_partner_operating_unit/models/res_users.py
```diff
@@ -18,16 +18,17 @@ def create(self, vals):
...
     @api.multi
     def write(self, vals):
-        res = super().write(vals)
         if vals.get('default_operating_unit_id'):
             # Add the new OU
             self.partner_id.operating_unit_ids = \
-                [(4, res.default_operating_unit_id.id)]
-        return res
+                [(4, vals.get('default_operating_unit_id'))]
+        return super().write(vals)
```
max3903, line 26: @osi-scampbell I would still write the user first and then update the parner with:
```
    @api.multi
    def write(self, vals):
        res = super().write(vals)
        if vals.get('default_operating_unit_id'):
            # Add the new OU
            self.partner_id.operating_unit_ids = \
                [(4, vals.get('default_operating_unit_id'))]
        return res
```

osi-scampbell: @max3903 If we do that then the check_partner_operating_unit() method runs and throws that error because the we have not written the default ou to the operating_unit_ids first. So we update the partner, then update the user.

max3903 (approved):

bodedra (approved):
Code LGTM :+1:

AaronHForgeFlow (commented):
Could you check CI before merging? It seems tests in purchase_requests makes sense.

osi-scampbell:
@AaronHForgeFlow I agree, I will write unit tests for this module. Going through CI it looks like there are multiple tests failing accross multiple modules, I will investigate those while writing test cases for this purchase_operating_unit

max3903:
@osi-scampbell Can you get the Travis error fixed?

murtuzasaleh (approved):
LGTM

murtuzasaleh:
@AaronHForgeFlow Can you please review this PR?

PR #240, opened 2020-01-17
[IMP] purchase_operating_unit: Enable Multi Picking/Warehouse on PO's

osi-scampbell:
Do not throw the UserError if we have a warehouse that does not have an operating unit (shared warehouse)

### DIFF
--- a/purchase_operating_unit/models/purchase_order.py
+++ b/purchase_operating_unit/models/purchase_order.py
@@ -89,9 +89,10 @@ class PurchaseOrder(models.Model):
     def _onchange_operating_unit_id(self):
         type_obj = self.env['stock.picking.type']
         if self.operating_unit_id:
-            types = type_obj.search([('code', '=', 'incoming'),
-                                     ('warehouse_id.operating_unit_id', '=',
-                                      self.operating_unit_id.id)])
+            types = type_obj.\
+                search([('code', '=', 'incoming'),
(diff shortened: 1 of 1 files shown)


purchase_operating_unit/models/purchase_order.py
```diff
@@ -92,6 +92,10 @@ def _onchange_operating_unit_id(self):
             types = type_obj.search([('code', '=', 'incoming'),
                                      ('warehouse_id.operating_unit_id', '=',
                                       self.operating_unit_id.id)])
+            if not types:
+                types = type_obj.search([('code', '=', 'incoming'),
+                                     ('warehouse_id.operating_unit_id', '=',
+                                      False)])
```
max3903: @osi-scampbell what about using:
```
types = type_obj.search([
    ('code', '=', 'incoming'),
    ('warehouse_id.operating_unit_id', 'in', (self.operating_unit_id.id, False)
])
```

max3903: @osi-scampbell I meant to merge this search with the one before (line 92)...

max3903 (approved):

AaronHForgeFlow (approved):

bodedra (approved):

AaronHForgeFlow (commented):
It is ok to me as soon lint checks are fixed.

PR #249, opened 2020-02-12
[12.0][FIX] account_operating_unit: Fix refund creation

jcoux:
When create a refund from an invoice,
we want to report the operating unit from original invoice to refund.


AaronHForgeFlow (approved):
:+1:  LGTM

Perhaps it is happening similar issue when using the reconciling widget. I created a PR in v10 sometime ago. I will check in this version too: https://github.com/OCA/operating-unit/pull/190

i-vyshnevska (approved):

MiquelRForgeFlow (approved):

PR #260, opened 2020-03-07
[IMP] stock_operating_unit: Group By Error + Location OnChange

osi-scampbell:
Just a few quality of life improvements for Stock OU's. If the OU's don't line up for the Location then it halts the workflow with  “The stock moves must be related to a location that belongs to the requesting OU”. This ensures those stay in line. The Group By on Pickings throws an error with that domain so we remove it.


max3903 (approved):

stock_operating_unit/model/product_product.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2020 Open Source Integrators
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo import _, api, models
```
AaronHForgeFlow: Travis complaining about the `_`

stock_operating_unit/model/product_product.py
```diff
@@ -0,0 +1,32 @@
...
+                search(domain or [], offset=offset, limit=limit, order=order)
+        else:
+            records = self.\
+                search(domain or [], offset=offset, limit=limit, order=order)
+        if not records:
+            return []
+        if fields and fields == ['id']:
+            # shortcut read if we only want the ids
+            return [{'id': record.id} for record in records]
+        if 'active_test' in self._context:
```
AaronHForgeFlow: can you explain a bit what is this use case?

AaronHForgeFlow (commented):
Thanks for the improvements, just a couple of comments

AaronHForgeFlow (approved):
:+1: Please squash before merge

PR #267, opened 2020-03-23
[IMP] stock_operating_unit (Update Record Rule)

patrickrwilson:
If the picking type is not part of a particular warehouse (like dropshipping) then errors will happen on purchase orders if dropshipping is chosen. This modifies the record rule to allow dropshipping (and other general, non-warehouse specific picking types) to work again on purchase orders.

### DIFF
--- a/stock_operating_unit/security/stock_security.xml
+++ b/stock_operating_unit/security/stock_security.xml
@@ -79,7 +79,8 @@
 
     <record id="ir_rule_stock_picking_allowed_picking_type_operating_units" model="ir.rule">
         <field name="model_id" ref="stock.model_stock_picking"/>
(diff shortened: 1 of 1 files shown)


patrickrwilson:
Ping @max3903 @osi-scampbell

max3903 (approved):

AaronHForgeFlow (approved):
:+1:

AaronHForgeFlow:
I remember there was a similar fix in another PR, do you know about that?

max3903:
@AaronHForgeFlow I don't remember...

patrickrwilson:
@AaronHForgeFlow I believe there was a same fix but was for another rule. This is the same fix that is needed for this rule too.

PR #268, opened 2020-03-23
[ADD] hr_operating_unit

patrickrwilson:
Adds a module called hr_operating_unit to show the related users operating units and default operating unit on the employee records.


patrickrwilson:
Ping @max3903 @osi-scampbell

osi-scampbell:
@patrickrwilson A couple of things: First, when making a new module in OCA we want to make sure we include this: 'Odoo Community Association (OCA)' in the authors section of the __manifest__.py or Travis will be red.

Second, this module is purely for reporting purposes, so we can see the OU's on the employee record instead of having to click to the partner record correct? Should we be adding record rules such as employees can only view other employees who have common OU's?

meigallodixital:
There was one PR in process: https://github.com/OCA/operating-unit/pull/194

patrickrwilson:
@osi-scampbell The OCA reference is already in the author section but i think it's complaining because i was missing the readme file which I just pushed.

@meigallodixital sorry, I didn't do my job and looked beforehand to see if there was anything that already existed. Your PR includes enabling the OU's on employee records which @osi-scampbell is suggesting while mine just adds related fields on the Employee form to show the related OU's on the User. The reason I added those fields was so that our HR department personnel can see the OU's on that are related to the user accounts without giving access to the users access or settings access.

It looks like that PR was stalled for a while but how would you want to proceed? We can either add the few items in mine into yours, or visa versa. It doesn't matter to me.

osi-scampbell:
@patrickrwilson I think we can have @meigallodixital PR for OCA, but maybe just add a few related fields in our curstom-addons repo? Might need @max3903 to weigh in on that decision

max3903:
Agree with @osi-scampbell. The user field on employee is not required, so let's use @meigallodixital's version and convert the fields to related ones privately.

patrickrwilson:
Ok, closing this and will re-add these to our private module.

Issue #270, opened 2020-03-25
[12.0] Default Operating Unit Res User Multi-Company


osi-scampbell:
We previously tried to make the default_operating_unit_id a company dependant field but upon further testing it was discovered that this feature created issues. For more details see the PR where we roll back this change here: #247. The requirement that we have not satisfied yet is for users who may belong to one OU in Company 1 and another in Company 2. Currently, the default_operating_unit_id stays the same for users regardless of which company they are logged in under.

max3903:
ping @AaronHForgeFlow

AaronHForgeFlow:
It is ok to me to add that feature as long we solve all the incompatibilities that comes up. Many PRs were red because of that and I don't like to break anything.

osi-scampbell:
@AaronHForgeFlow Ahhh I see, so maybe the solution is we open a [WIP] PR to see where the repo has its incompatabilites, then make PR's to change those modules affected?

AaronHForgeFlow:
> @AaronHForgeFlow Ahhh I see, so maybe the solution is we open a [WIP] PR to see where the repo has its incompatabilites, then make PR's to change those modules affected?

Yes, agree with that. For some reason the PR that added this feature  was green, but then after the merge, some PRs were failing for unknown reasons. I remember a fix in module stock operating unit  that was causing an unrelated error in purchase_request_operating_unit, and the logs were not clear. Let's see now.

PR #272, opened 2020-03-27
[FIX] operating_unit: Default OU company-dependent

max3903:
#270


murtuzasaleh:
#306 It will solve UT Case.

flotho:
Does this PR still in WIP ?
Any review needed ?

Issue #258, opened 2020-02-26
[12.0] Performance issue with res_partner_operating_unit


max3903:
After installing and configuring this module, the contacts list is pretty slow (~45 sec with 115 k records).

From Odoo's support comment below, a record rule on a many2many field is not good a idea:

> Many2many fields security rules on big tables are indeed sub-optimal compared to a subselect or join done in SQL directly. This is unfortunate for your case, but this is the expected behaviour of Odoo at this time. Improving this is something we keep in mind for the future, but using joins or more efficient subqueries comes with its own problems, and at the moment we have no real incentive to make this a priority - security rules based on m2m fields are not present in standard modules for a reason; the same way that searching on big tables by followers or by tags could be rather lengthy. It's not really a problem when it's a search - for security rules, since it's added to every query, the impact is indeed more important.

Any idea or suggestion? I want to make sure we are not forgetting a good solution.

Thanks

max3903:
CC: @moylop260

moylop260:
If I understood fine the following line is the performance issue:
 - https://github.com/OCA/operating-unit/blob/040b17de8b1e88cb74d77a9478e8fddccb3d8547/res_partner_operating_unit/security/res_partner_security.xml#L10

I don't know what is the best solution in this moment.
Fast solution in my mind (without thinking too much), maybe, you like to try to make a field in `res.user` to get all valid partner ids for that user (using a faster query directly)

In order to use:
 - ` <field name="domain_force">[('id', 'in', user.partner_allowed_by_ou_ids.ids)]</field>`

Offtopic, You need to clear caches after `partner.operating_unit_ids` field changes, since that [_compute_domain](https://github.com/odoo/odoo/blob/87b127efed97c1f6fad18d6ee6b5924a9b2c70f2/odoo/addons/base/models/ir_rule.py#L71-L77) method has `LRU` cache, but it is a dynamic value, if you don't clear caches so changes in the fields won't be detected.

max3903:
Thanks @moylop260 I think we'll try the other way around (storing the users on the partners) as we have 200+ users and 100+ k partners.

max3903:
Issue persists with both options.

Trying the following option:
* Remove the record rule
* Filter on the OUs in `name_search`

IJOL:
[...]

PR #274, opened 2020-04-01
[FIX] res_partner_operating_unit

max3903:
#258 

Having the field required simplifies the record rule and solves the performance issue.

### DIFF
--- a/res_partner_operating_unit/hooks.py
+++ b/res_partner_operating_unit/hooks.py
@@ -1,0 +1,19 @@
+# Copyright (C) 2020 Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+
+def pre_init_hook(cr):
+    # Add new table and columns to hold values
+    cr.execute("""
+        CREATE TABLE operating_unit_partner_rel (
+            partner_id INTEGER NOT NULL
+                REFERENCES res_partner(id) ON DELETE CASCADE,
+            operating_unit_id INTEGER NOT NULL
+                REFERENCES operating_unit(id) ON DELETE CASCADE);
+    """)
+    # Add the values
+    cr.execute("""
+        INSERT INTO operating_unit_partner_rel
+            (partner_id, operating_unit_id)
+        SELECT id, 1 FROM res_partner;
+    """)
--- a/res_partner_operating_unit/__init__.py
+++ b/res_partner_operating_unit/__init__.py
@@ -1,2 +1,3 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 from . import models
+from .hooks import pre_init_hook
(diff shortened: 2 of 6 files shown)


res_partner_operating_unit/hooks.py
```diff
@@ -0,0 +1,20 @@
...
+                REFERENCES res_partner(id) ON DELETE CASCADE,
+            operating_unit_id INTEGER NOT NULL
+                REFERENCES operating_unit(id) ON DELETE CASCADE,
+        );
+    """)
+    # Add the values
+    cr.execute("""
+        INSERT INTO operating_unit_partner_rel
+            (partner_id, operating_unit_id)
+        SELECT id, 1 FROM res_partner;
```
AaronHForgeFlow, line 18: Why operating_unit is always the first?

max3903: I don't know...

AaronHForgeFlow: I am not too familiar with the module. However, if we add partners to all OUs on the installation, all partners will belong to all OU, won't they? when the module is not installed all partners belong to all OU. From my point of view, if we include by default all partners to all OUs will be consistent. So I'd loop over all OUs assigning partners. 

Does that look correct to you?

max3903: Humm not really. The field is required so we need to set a default value. The only value we have is the "Main OU" created by the `operating_unit` module.

The hook and the migration script only set one OU ("Main OU") to all the partners (hooks) or all the partners with no OU (migration script).

AaronHForgeFlow: hmm I see. Ok from my side go ahead with this. The only doubt is if there is a chance that id=1 was delete from some reason before the module is installed. Perhaps we can do a select before.

patrickrwilson (approved):

AaronHForgeFlow:
LGTM, could you check travis?

res_partner_operating_unit/hooks.py
BT-rmartin: @max3903  Why are the migration script and hook needed?
Isn't the default method in operating_unit_ids working and assigning the user default_operating_unit_id?

max3903: @BT-rmartin Defaults works when you create a new record. It will not update the existing ones. 

The hook is executed when the module is installed and set the Main OU on all existing partners.

The migration script is executed when the module is upgraded and set the Main OU on existing partners with no OU.

BT-rmartin: @max3903 I did check and defaults are applied also to existing data. Maybe not m2m, but for basic fields they are. 
From my understanding, default is applied to both new and existing data. 
pre-init/post-init hooks only when installing for existing data.
migration script only when upgrading for existing data. 
In fact I see in another PR https://github.com/OCA/operating-unit/pull/225 that you removed the hook. Is this really necessary here?

gdgellatly:
I don't think this should be merged. It is perfectly valid to have no operating unit on a partner such that it can be seen by all operating units. Just as with company.

Employee records, Company records etc.

If anything it makes more sense to have operating unit as a many2one field on partners in particular, My most common use cases are all or one.

PR #283, opened 2020-05-06
[13.0][IMP] Add some feature in account_operating_unit module

newtratip:
- _check_move_operating_unit for all move types
- When you add invoice line operating unit should show in journal item immediately (not after save)
- Check company in move and operating unit must equal
- Filter analytic account in move line with operating unit on move
- Add operating unit in invoice's tree view
- Add search and group by in invoice

### DIFF
--- a/account_operating_unit/models/account_move.py
+++ b/account_operating_unit/models/account_move.py
@@ -224,9 +224,24 @@ class AccountMove(models.Model):
                 and move.operating_unit_id
                 and move.operating_unit_id != move.journal_id.operating_unit_id
             ):
-                raise UserError(
-                    _("The OU in the Move and in Journal must be the same.")
-                )
+                # Change journal_id if create move from other model. e.g., sale.order
+                if (
+                    move._context.get("active_model")
+                    and move._context.get("active_model") != "account.move"
+                ):
+                    move._onchange_operating_unit()
+                    if (
+                        move.journal_id.operating_unit_id
+                        and move.operating_unit_id
+                        and move.operating_unit_id != move.journal_id.operating_unit_id
+                    ):
+                        raise UserError(
+                            _("The OU in the Move and in Journal must be the same.")
+                        )
+                else:
+                    raise UserError(
+                        _("The OU in the Move and in Journal must be the same.")
+                    )
         return True
 
     @api.constrains("operating_unit_id", "company_id")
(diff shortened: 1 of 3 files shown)


newtratip:
Hello @alan196, Can you help review this ?

AaronHForgeFlow (commented):
Code review LGTM

alan196 (approved):
:+1: Code review

newtratip:
Reference https://github.com/OCA/operating-unit/issues/284
I change journal_id on constrain if pass active_model in move's context
Selection_493

This is work for me, @AaronHForgeFlow Can you review this again ?

Thanks !!

AaronHForgeFlow:
> Reference #284
> I change journal_id on constrain if pass active_model in move's context
> Selection_493
> 
> This is work for me, @AaronHForgeFlow Can you review this again ?
> 
> Thanks !!

@newtratip thanks for the fix. In what cases there is no 'active_model' in the context? only for invoices?

newtratip:
@AaronHForgeFlow, I think only invoice no active_model. Can you suggest me ?

account_operating_unit/models/account_move.py
```diff
@@ -223,8 +223,36 @@ def _check_journal_operating_unit(self):
                 move.journal_id.operating_unit_id
                 and move.operating_unit_id
                 and move.operating_unit_id != move.journal_id.operating_unit_id
+            ):
+                # Change journal_id if create move from other model. e.g., sale.order
+                if move._context.get("active_model"):
```
AaronHForgeFlow:
```suggestion
                if move._context.get("active_model") and move._context.get("active_model") != 'account.move':
```

AaronHForgeFlow: I think this is safer but in any case your code was working. Thanks.

newtratip: Why do you use `move._context.get("active_model") == 'account.move'`. I think `move._context.get("active_model") != 'account.move'` because If we created invoice from other model, active_model will be that model. This is change journal for not created from account.move directly.

WDYT, Thank for your suggest :)

AaronHForgeFlow: Yes sorry, I meant != . But in any case it is not big deal. I tested functional and it works.

AaronHForgeFlow (approved):
Functional test :+1:

PR #286, opened 2020-05-14
[12.0][ADD]WIP pos_operating_unit

### LINKED ISSUE #155: [RFC] OU in POS (12.0 ideally)
Hi,
Is there any development/intention related for OU in POS? I could not find any in this repo but maybe some other repos might be looked into?

If not we might plan to use OU in one of our project and would need OU in POS. AFAICS, we would need to add the field/ACL in the following objects:
* pos.session
* pos.order

I am wondering about the payment methods and POS Product Categories.
Besides we will need to be careful with:
* the journals set up
* potential conflicts with accounting OU?

Any other idea/feedback?

flotho:
This module is clearly inspired by sale_operating_unit.
This fix #155 

Please add WIP on it and feel free to comment, more commits will arrive


flotho:
@elicoidal , @jacobus101,
Your comments will be appreciated

luistriana28:
👍

PR #287, opened 2020-05-14
Reduce the width of dependency from analytic_operating_unit to operat…

glitchov:
Analytic has nothing to do with this module but operating_unit well.


luc-demeyer (approved):

AaronHForgeFlow (approved):
The account module depends on analytic module. That is why the dependency was included at first. But technically speaking it is not necessary. I guess that should be enough to remove the dependency.

pedrobaeza:
There has been a problem with merge bot, but it's fixed now, so you can relaunch.

PR #296, opened 2020-05-26
[ADD] stock_warehouse_orderpoint: added functionality of operating unit

jesus01x:
### Proposed changes

This module allow use operating unit in stock.warehouse.orderpoint

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated necessary documentation

### Further comments

Aditional comments

### DIFF
--- a/stock_operating_unit/README.rst
+++ b/stock_operating_unit/README.rst
@@ -31,6 +31,7 @@ This module introduces the following features:
 - Adds the operating unit to the Stock Location.
 - Adds the requesting operating unit to stock pickings.
 - Implements user's security access rules.
+- Adds the operating unit to the Stock Warehouse Orderpoint.
 
 **Table of contents**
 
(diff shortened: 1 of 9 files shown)


AaronHForgeFlow (commented):
Hi! Could you let us know why is the operating unit in the orderpoint needed? Orderpoints are actually linked to an operating unit because the warehouse and the location has already an operating unit. :thinking:

alan196:
> Hi! Could you let us know why is the operating unit in the orderpoint needed? Orderpoints are actually linked to an operating unit because the warehouse and the location has already an operating unit. 🤔

This change will make the order points visible only to the operating unit the users are using. If a user see an orderpoint of another OU it will confuse them

PR #297, opened 2020-05-29
[13.0][MIG]product_operating_unit

Dangar-Serpentcs:
Migration product_operating_unit into v13.


alan196:
Please follow instructions of https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 and check travis

BT-fgarbely:
Hello @Dangar-Serpentcs 
What's the status of this Pull Request? We need this module as well, and it would be great to have it migrated as soon as possible.
Is the migration finished from your side an ready to be reviewed?

Dangar-Serpentcs:
> Hello @Dangar-Serpentcs
> What's the status of this Pull Request? We need this module as well, and it would be great to have it migrated as soon as possible.
> Is the migration finished from your side an ready to be reviewed?

Yes, @BT-fgarbely for the product_operating_unit module migration done from my end.

Hello @AaronHForgeFlow can you please review this Merge request.

nikul-serpentcs (approved):
Code Review LGTM :+1:

nikul-serpentcs:
@BT-fgarbely @AaronHForgeFlow Can you please review this PR?

AaronHForgeFlow:
@Nikul-Chaudhary I am afraid I am not familiar with this module. I merged the fix on this module in v12 because it was a blocking bug, but, for the whole module I prefer actual users to check.

Dangar-Serpentcs:
Hello @AaronHForgeFlow We need this module functionality.

AaronHForgeFlow:
Hi @Dangar-Serpentcs  I will not block this because it is in v12 but I will not review because the reason commented in previous comment. Regards.

luc-demeyer (changes requested):
I think this module suffers from the same issues as why I proposed https://github.com/OCA/operating-unit/pull/350

Adding complex m2m relationships between products and operating units with record rules allowing to share products amongst a subset of operating units will cause perfornance degradation when having large number of product records. It will also be hard to manage.

I think we should do the same as what I propose for the partners.
Create a basic 'product_operating_unit' module that adds a m2o from product record to operating_unit and a record rule allowing to isolate products (but make them shared by default).

We should imho rename this module to product_multi_operating_unit for those companies who need to share products amongs a subset of their operating units. But I think that except renaming also a number of changes are required to this module to make this approach workable (e.g. imho default should be shared and not OUs of user, default function should be easily inheritable, hence called via 'lambda self' in the fields definition, constraints need to be carefully reviewed and tested, category has super on 'write' but not on create, ...).

PR #298, opened 2020-05-29
[IMP] stock_operating_unit: Check Destination Location OU is Set

osi-scampbell:
Small change here, if an OU is False if a source location or a destination location does not have an OU then that means anything can come from/to those locations. We are consistent with source location but we are not checking to see if destination OU is set. This causes an error when users are trying to push to a location with no OU set and still getting this error thrown.


max3903 (approved):

murtuzasaleh (approved):

Chandresh-SerpentCS (approved):
LGTM

PR #305, opened 2020-07-16
[12.0][FIX] res_partner complete filter also in name search

andreampiovesana:
if you start to insert a partner you visualize all, with this you immediately filter partner by operating-unit


gregoril (approved):
Tested, it works correctly

patrickrwilson (approved):
Tested and works as intended.

marcelsavegnago (changes requested):
Please check travis error

PR #308, opened 2020-08-06
[13.0][MIG] res_partner_operating_unit

Dangar-Serpentcs:
Migration res_partner_operating_unit into v13.


Dangar-Serpentcs:
Hello @AaronHForgeFlow  Can you please review this #PR, Here this #PR Travis failed because of the test_hr_expense_operating_unit module test case.
Screenshot from 2020-08-06 18-25-03

luc-demeyer:
@HirenDangar 
Can you rebase your PR.
This should fix the travis error.

Dangar-Serpentcs:
> @HirenDangar
> Can you rebase your PR.
> This should fix the travis error.

Hello, @luc-demeyer travis fixed now.

luc-demeyer (changes requested):
This module doesn't work at all. Needs refactoring and imho in the meanwhile should be set to installable False (and the same for the 12.0 version of this module.

luc-demeyer:
I installed this module and tried to get it to work and can only come to the conclusion that we should put the 12.0 module to 'installable: False' and not merge this 13.0 port unless someone spends time to refactor it.

More detailled explanation:

https://github.com/OCA/operating-unit/blob/6a1b2e892c08437b5642acca73bcaa3a06bd71b2/res_partner_operating_unit/models/res_users.py#L25

This line is not covered by unit test, but a basic test via the User Interface gives a stack trace.
Looking to the code:

    @api.multi
    def write(self, vals):
        res = super().write(vals)
        if vals.get('default_operating_unit_id'):
            # Add the new OU
            self.partner_id.operating_unit_ids = \
                [(4, res.default_operating_unit_id.id)]
        return res

res is a boolean and the code does a "res.default_operation_unit_id.id".
Hence my conclusion that this module is not used in this form in a production environment and hence we can put to uninstallable without impact.

A bit further in the file:

https://github.com/OCA/operating-unit/blob/6a1b2e892c08437b5642acca73bcaa3a06bd71b2/res_partner_operating_unit/models/res_users.py#L28:

    @api.constrains('partner_id.operating_unit_ids',

@api.constrains only works on simple fields not dotted fields (partner_id.operating_unit_ids hence this constraint doesnt work.
There is also something strange in the logic here since the create/write methods add the default_operating_unit_id to the partner operating_unit_ids but we check this in the constrains and a constrains method is executed before we hit the create/write methods. If someone has time/budget to refactor this module than this logic must be fixed.

There is also a problem with the module name:  res_partner_operating_unit
I started to look to this module because I expected a basic addition of operating_unit_id to partner records but than I see that this module has been made to allow sharing partners amongst different operating units, leading to many2many on res.partner and the fact that record rules can not be used for performance reasons.
If someone refactors this module than I propose that it comes back on another name:
res_partner_multi_operating_unit
and the README must be adapted to document what this module does, now it says: Adds the Operating Unit (OU) to res partner.

@max3903 
Maxime, I see that you have been involved in the M2M performance discussion when this module was created.
Do you agree with my analysis and an eventual 'reborn' of this module under another name ?

Concerning the basic functionality of adding an operating_unit on res.partner:
I think we can simply add this to the base operating_unit module: only a few lines of code and a record rule (and no default function, hence by default partners are shared amongst operating units).
I'll make a PR for this.

Dangar-Serpentcs:
Hello, @luc-demeyer already reported the issue. https://github.com/OCA/operating-unit/pull/336/files

luc-demeyer:
cf. https://github.com/OCA/operating-unit/pull/349

andreampiovesana (approved):
go

PR #309, opened 2020-08-16
[13.0] [MIG] project_operating_unit

nikul-serpentcs:


AaronHForgeFlow:
Same as v12. IMHO as long as project depends on analytic, this module should also depend on analytic _operating_unit. But, in that case, we need to make sure there is not discrepancies between the operating unit of the project with the operating unit of the analytic account.

PR #312, opened 2020-08-19
[13.0]fix operating_unit record rules

luc-demeyer:
In the 12.0 migration of this module the record rules got changed without any explanation why.
Imho this change is not required and is also confusing since the module now adds group level record rules but the group is a technical usability group to hide the operating unit fields from the User Interface.

I have done a couple of tests with the record rules coming with this PR and I didn't find any issue (and also didn't expect any issue since the 'old' record rules have worked correctly up to oca/operating_unit 11.0).

Maybe I am missing something but if not than I think we should restore the record rules to how it was up to the 12.0 port.

### DIFF
--- a/operating_unit/models/res_users.py
+++ b/operating_unit/models/res_users.py
@@ -24,16 +24,43 @@ class ResUsers(models.Model):
     def _default_operating_units(self):
         return self._default_operating_unit()
 
-    operating_unit_ids = fields.Many2many(
-        "operating.unit",
-        "operating_unit_users_rel",
-        "user_id",
-        "operating_unit_id",
-        "Operating Units",
+    operating_unit_ids = fields.One2many(
+        comodel_name="operating.unit",
+        compute="_compute_operating_unit_ids",
+        inverse="_inverse_operating_unit_ids",
+    )
+    assigned_operating_unit_ids = fields.Many2many(
+        comodel_name="operating.unit",
+        relation="operating_unit_users_rel",
+        column1="user_id",
+        column2="operating_unit_id",
+        string="Operating Units",
         default=lambda self: self._default_operating_units(),
     )
+
     default_operating_unit_id = fields.Many2one(
-        "operating.unit",
-        "Default Operating Unit",
+        comodel_name="operating.unit",
+        string="Default Operating Unit",
         default=lambda self: self._default_operating_unit(),
     )
+
+    @api.depends("groups_id", "assigned_operating_unit_ids")
+    def _compute_operating_unit_ids(self):
+        for user in self:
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
I think that change was done in v13, I don't remember why and need to recheck.

luc-demeyer:
@AaronHForgeFlow 
I looked into the git history. This change came with 12.0 migration.

I think I understand why this change has been done:

The "Operating Unit Manager" group has been introduced as part of the 12.0 port and users of this group can manage all operating units even if they don't have them in their "operating_unit_ids" via the "[(1, '=', 1)]" domain.

But in order for this rule to work the previous "ir_rule_operating_unit_allowed_operating_units" global rule had to be turned into a group rule.

Hence all at a sudden a usabilty group which was there in older versions to hide OU from the User Interface has now turned into a group which becomes necessary for all OU users.

This looks like a breaking change to me.

I also find the "Operating Unit Manager" group confusing.
Having a user without Operating Units who can manage all of them is not very intuitive.

We also don't need it.
With the ACLs every user in this group can already create new Operating Units.
The only thing that is missing is that a OU created by e.g. the admin will not be added to the users of the "Operating Unit Manager" group.
We could easily fix this via the "create" method on the OperatingUnit Class.

I propose to extend this PR further into this direction but I think it would be usefull to involve a few more member of the "Operating Unit" steering commitee in this discussion so that we all agree on the way forward before doing the work.

I didn't find the list of Operating Unit PSC members on the OCA web page hence can you involve your PSC members ?

AaronHForgeFlow:
I remember having issues because the operating unit manager was not able to create or manage operating units. Indeed, in previous versions it was the same, but the user root was over-passing security checks and that is why there was no issues. In order to solve that, as you said ir_rule_operating_unit_allowed_operating_units was modified to be a group rule. I did not like that because operating unit rules should be global rules, but at that time it was the cleanest solution found. I am not sure if the solution proposed here, adding an auxiliary field is appropriate, but I think I prefer this than what we have now.

In any case I am having issues after creating a new operating unit in the system using admin user and trying to access admin user form view on runbot.

luc-demeyer:
@AaronHForgeFlow 
can you check if you still have issues after fix https://github.com/OCA/operating-unit/pull/312/commits/0a13526a652c6f85b6f36c320ee707413364b1e1 ?

AaronHForgeFlow (approved):
:+1: Seems to be working for me. Will wait for more reviewers before merge as this is a major change.

alan196 (approved):

AaronHForgeFlow:
@luc-demeyer could you squash commits before merging? Thanks in advance

PR #317, opened 2020-09-17
[FIX] proper calculation of request_line_ids from active_model

kh-badep:
the default_get method supposes that 'active_ids' is the id of the purchase request lines. Which is not the case when the wizard is called from the purchase request. This PR fixes it by testing on 'active_model' and making proper calculations.
See: https://github.com/OCA/purchase-workflow/commit/4fe36979f824d0b9b3680e5817bb524ab2eef5df#diff-e458197a92be2847dc330973def5bc39


kh-badep:
Replaced by: https://github.com/OCA/operating-unit/pull/318

PR #318, opened 2020-09-17
[FIX] proper calculation of request_line_ids from active_model

kh-badep:
the default_get method supposes that 'active_ids' is the id of the purchase request lines. Which is not the case when the wizard is called from the purchase request. This PR fixes it by testing on 'active_model' and making proper calculations.
See: OCA/purchase-workflow@4fe3697#diff-e458197a92be2847dc330973def5bc39

### DIFF
--- a/purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
@@ -22,7 +22,16 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         res = super(PurchaseRequestLineMakePurchaseOrder, self).\
             default_get(fields)
         request_line_obj = self.env['purchase.request.line']
-        request_line_ids = self._context.get('active_ids', [])
+        active_model = self.env.context.get('active_model', False)
+        request_line_ids = []
+        if active_model == 'purchase.request.line':
+            request_line_ids += self.env.context.get('active_ids', [])
+        elif active_model == 'purchase.request':
+            request_ids = self.env.context.get('active_ids', False)
+            request_line_ids += self.env[active_model].browse(
+                request_ids).mapped('line_ids.id')
+        if not request_line_ids:
+            return res
         operating_unit_id = False
         for line in request_line_obj.browse(request_line_ids):
             line_operating_unit_id = line.request_id.operating_unit_id \


purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -21,10 +21,18 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
     def default_get(self, fields):
         res = super(PurchaseRequestLineMakePurchaseOrder, self).\
             default_get(fields)
-        request_line_obj = self.env['purchase.request.line']
-        request_line_ids = self._context.get('active_ids', [])
+        active_model = self.env.context.get('active_model', False)
+        request_line_ids = []
```
AaronHForgeFlow, line 26: I think it is not necessary to predefine this variable as long there is no lop, you can define the variable inside de if statament. don't you?

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -21,10 +21,18 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
     def default_get(self, fields):
         res = super(PurchaseRequestLineMakePurchaseOrder, self).\
             default_get(fields)
-        request_line_obj = self.env['purchase.request.line']
```
AaronHForgeFlow, line 24: Ok this variable may only be used once but as long as the purpose of the pull request is different I would prefer to skip it for now, the relevant change is the if statement

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
kh-badep: Yes I tought so too, but I opted to stay on the same logic of the original PR for the purchase_request module: https://github.com/OCA/purchase-workflow/commit/4fe3697#diff-e458197a92be2847dc330973def5bc39R99

Though I do think it is unnecessary in both cases.

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
kh-badep: OK. Do you want me to make another PR without it?

AaronHForgeFlow: You can do the change in this PR but I will not block the PR for this. It is just for reviewers to focus on what the PR is about.

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
AaronHForgeFlow: I see it was there before. But it will be an improvement,  you could take advantage of this PR and change it :) Not blocking anyway

kh-badep: Upon further investigation I think it was done to simplify the exit case when no request lines are defined OR when there is a bad context here: https://github.com/OCA/purchase-workflow/commit/4fe3697#diff-e458197a92be2847dc330973def5bc39R106

Otherwise we would have to individually test if there are some request lines or if the context is properly configured. Or wrap the whole logic after on an `if request_line_ids` statement, which will further make the code complex

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
kh-badep: Done

kh-badep (commented):
revert refactoring of request_line_obj

AaronHForgeFlow (approved):
Thank you @metamasterplay  for attending the comments. Just please squash the commits. From my side it is ok. :+1:

purchase_request_operating_unit/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -22,7 +22,16 @@ def default_get(self, fields):
         res = super(PurchaseRequestLineMakePurchaseOrder, self).\
```
astirpe, line 22: I think the entire method could be simplified and made more efficient. Something like this:

```python
    @api.model
    def default_get(self, fields):
        res = super().default_get(fields)
        requests = self.env['purchase.request']
        active_model = self.env.context.get('active_model', False)
        active_ids = self.env.context.get('active_ids', [])
        active_recordset = self.env[active_model].browse(active_ids)
        if active_model == 'purchase.request.line':
            requests |= active_recordset.mapped('request_id')
        elif active_model == 'purchase.request':
            requests |= active_recordset
        operating_units = requests.mapped("operating_unit_id")
        if len(operating_units) >= 2:
            raise ValidationError(_('Could not process ! '
                                    'You have to select lines '
                                    'from the same operating unit.'))
        res['operating_unit_id'] = operating_units.id
        return res
```

AaronHForgeFlow: I think it is a good opportunity to fix this method now. And do a forward port to 13.0.

kh-badep: Sounds good for me too. Just let me know if I need to do something on my side, not an expert on GitHub per se.

kh-badep:
Hello @AaronHForgeFlow . I think I've managed to squash the commits. Let me know if there's anything else left to do on my side.

AaronHForgeFlow:
@metamasterplay  just need reviewers

@astirpe  does this look correct?

kh-badep:
Hello @astirpe  @AaronHForgeFlow. 

Just a reminder for this PR

astirpe:
@kh-badep in my previous comment I was proposing a refactoring of the `default_get()` method. Are you going to work on it?

PR #319, opened 2020-09-21
[ADD] operating_unit_hierarchical

hbrunn:
This module allows to configure a hierarchy of operating units.

Parent OUs behave as if they contain their child OUs, so if a user has access to some OU, its children are accessible too.


AaronHForgeFlow:
This is a nice feature.

Will the field operating_unit_ids in the user include all the children operating units? Otherwise users will not have access to records they are supposed to, for example, in the module sale_operating_unit when accessing sales order of a child operating unit.

hbrunn:
currently, this indeed must be done manually. I'm undecided whether the best approach here is making `res.user#operation_unit_ids` a computed field in the module that looks at some other field where the manually added OUs live and generates all the children for them, override searching for `in` during `ir.rule` evaluation, or change all the rules to something like `in user._get_accessible_ous()` which any module can override as it sees fit. The latter might be the most versatile approach.

I'm looking into fixing CI

hbrunn:
tests should be fixed in https://github.com/OCA/operating-unit/pull/320

AaronHForgeFlow:
Tests should be fixed after rebase.

`change all the rules to something like in user._get_accessible_ous() which any module can override as it sees fit`

I agree this option is more flexible :)

hbrunn:
@AaronHForgeFlow I implemented that, and rebased

AaronHForgeFlow:
Hi @hbrunn  I am sorry for my previous message, but I have think about this and I think the best approach is not to override all the record rules. I think it is best to follow the same approach Odoo does for companies. I mean, if there is a new child operating unit and you need the users with access to the parent to access to the child too then the best option is to include the child into the operating_unit_ids of the users. Either programatically or manually. What I mean is, the problem of accessing the records belonging to a specific operating unit it is already solved and it works, and it is easier to maintain. 

In addition to that, there are many constraints in the different modules that uses the operating_unit_ids field of the user to be evaluated. Those constraints will still raise because those just check the operating_unit_ids field, without any hierarchy. Same as before, if we just treat the child operating unit as another different operating unit, then there is no issue and we can continue using the same code.

So the best thing here would be just adding the hierarchy without record rules, and then, if needed, adding the users to the newly created children operating units. That is a simple approach and Odoo uses similar thing for companies.

Regards.

ntsirintanis (approved):
It works.

PR #328, opened 2020-10-10
[14.0][MIG] migration of module operating_unit 

alan196:
### Proposed changes

I've migrated the module operating_unit and its functionalities from Odoo 13.0 to 14.0. I've checked if all unit tests are still running.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

PR #331, opened 2020-10-10
[14.0][MIG] migration of module analytic_operating_unit

alan196:
### Proposed changes

I've migrated the module analytic_operating_unit and its functionalities from Odoo 13.0 to 14.0. I've checked if all unit tests are still running.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

### Further comments

This module depends on the operating_unit. That means that there should be a migration of it to 14.0. Because there is already a PR for migration I've added the changes of the PR to my local environment to test my migration.

This PR depends on the following migration PR:

https://github.com/OCA/operating-unit/pull/328


alan196:
@kittiu could you review this 🙏

Saran440:
Functional Test :+1: 

I have question.
If I config following
Selection_004

Why It filter default OU? Should It filter by OU not default OU? or I mistake something.
Selection_005

kittiu:
LGTM too

kittiu (approved):

alan196:
> Functional Test 👍
> 
> I have question.
> If I config following
> Selection_004
> 
> Why It filter default OU? Should It filter by OU not default OU? or I mistake something.
> Selection_005

I see an issue here, but it's not from this module it's from module operating_unit, this field has a domain `[('user_ids', '=', uid)]` in the field, but I see that this field in module operating_unit is only defined when the Operating Unit is created and it will be only the user that creates the OU.

I think is better to change the domain to filter `[('id', 'in', user.assigned_operating_unit_ids.ids)]`

Saran440 (approved):

PR #339, opened 2020-11-11
[ADD] New module for Switching Multiple OP in Menu Bar

Vanneri:
This module provide easiness to switch between multiple operating units


Vanneri:
how to overcome the issue?,, i have no idea @max3903

Issue #340, opened 2020-11-16
Change default in purchase_operating_unit/models/purchase_order.py


Borruso:
I have problem when run 'Procurement: run scheduler'.
I suggest that the operating_unit_id and requesting_operating_unit_id depends on warehouse stock.

Example:
I have several operating units and I create a stock rule with warehouse2 but when I start scheduler it shows me the error 'Configuration error. The Quotation / Purchase Order and the Warehouse of picking type must belong to the same Operating Unit' because operating_unit_id and requesting_operating_unit_id have as default user's opertaing_unit and not warehouse's opertaing_unit

Suggest:

- Update _check_warehouse_operating_unit

`@api.constrains('operating_unit_id', 'picking_type_id')`
`    def _check_warehouse_operating_unit(self):`
`        for record in self:`
`            picking_type = record.picking_type_id`
`            if not record.picking_type_id:`
`                continue`
`            # OdooBot run scheduler for all operating units`
`            if record.env.user.id == 1:`
`                continue`
`            warehouse = picking_type.warehouse_id`
`            if (warehouse.operating_unit_id and`
`                    picking_type.warehouse_id and`
`                    picking_type.warehouse_id.operating_unit_id and`
`                    record.operating_unit_id and`
`                    warehouse.operating_unit_id != record.operating_unit_id):`
`                raise ValidationError(`
`                    _('Configuration error. The Quotation / Purchase Order '`
`                      'and the Warehouse of picking type must belong to the '`
`                      'same Operating Unit.')`
`                )`

- Add super on create function

`@api.model`
`    def create(self, vals):`
`        res = super(PurchaseOrder, self).create(vals)`
`        for purchase in res:`
`            picking_type = purchase.picking_type_id`
`            if picking_type:`
`                warehouse = picking_type.warehouse_id`
`                if purchase.env.user.id == 1 and warehouse.operating_unit_id:`
`                    res['operating_unit_id'] = warehouse.operating_unit_id.id`
`                    res['requesting_operating_unit_id'] = warehouse.operating_unit_id.id`
`        return res`

PR #341, opened 2020-11-16
[12.0] Procurement: run scheduler in error for OdooBot

Borruso:
I have several operating units and I create a stock rule with warehouse2 but when I start scheduler it shows me the error 'Configuration error. The Quotation / Purchase Order and the Warehouse of picking type must belong to the same Operating Unit' because operating_unit_id and requesting_operating_unit_id have as default user's opertaing_unit and not warehouse's opertaing_unit


Borruso:
Possible solution for this problem #340

MarcoCalcagni (approved):
ok, we are using it in production

PR #343, opened 2020-12-04
[14.0][MIG] purchase_operating_unit [WIP]

kittiu:
Depends

- [x] stock_operating_unit https://github.com/OCA/operating-unit/pull/344

Note: test locally ok.


alan196:
@kittiu this PR it's still on WIP?

kittiu:
Itself no. But it is waiting for stock_operating_unit

feketemihai (approved):
Tested locally with stock_operating_unit and it works fine.

Saran440 (approved):
Please update your travis.
Functional Test :+1:

Saran440:
Functional Test:
Error when you have many ou and select po without default ou.

My login:
2 ou (ou1, ou2)
default with ou1

step to error
1. Create PO and change Operating Unit and Requesting Operating Unit to ou2
2. Received Product > Create bills
3. it throw error `The operating unit of the purchase order must be the same as in the associated invoices.`

purchase_operating_unit/models/purchase_order.py
```diff
@@ -0,0 +1,128 @@
...
+                        "in the Purchase Order"
+                    )
+                )
+
+    @api.model
+    def _prepare_picking(self):
+        picking_vals = super(PurchaseOrder, self)._prepare_picking()
+        picking_vals["operating_unit_id"] = self.operating_unit_id.id
+        return picking_vals
+
```
Saran440, line 121: def action_create_invoice(self):
        ctx = self._context.copy()
        for order in self:
            ctx["default_operating_unit_id"] = order.operating_unit_id.id or False
        return super(PurchaseOrder, self.with_context(ctx)).action_create_invoice()


I think we should add context default ou to invoice for fix it.

PR #234, opened 2020-01-15
[IMP] stock_operating_unit: Multi Warehouse Access

osi-scampbell:
This PR will make warehouses that do not have an operating_unit_id accesible to all Operating Units

#237

PR #344, opened 2020-12-04
[14.0][MIG] stock_operating_unit

kittiu:
Standard migration, ready for review.


alan196 (approved):
Technical Review 👍

Saran440 (approved):
Functional Test :+1: 
Could you fixed travis?

kittiu:
I think runbot is failing without reason. It just stop.

feketemihai (approved):

gustavovalverde (approved):

gustavovalverde:
cc: @max3903 this seems ready enough. Does the codecov needs to be solved for merge?

stock_operating_unit/model/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+                    )
+                )
+
+    @api.constrains("operating_unit_id", "picking_type_id")
+    def _check_picking_type_operating_unit(self):
+        for rec in self:
+            warehouse = rec.picking_type_id.warehouse_id
+            if (
+                rec.picking_type_id
+                and rec.operating_unit_id
```
Saran440, line 48:
```suggestion
                and rec.operating_unit_id
                and warehouse.operating_unit_id
```

This function should add domain for case warehouse don't have operating unit.
In my case:
Warehouse OU : False
User OU: Main OU

Step to test:
1. Create warehouse without OU.
2. Create PO (Main OU) and select product type Storable Product
3. Click Confirm button
4. It throw error because warehouse ou != my ou

Is it can receipt stock to warehouse?

-----------------------------

Edited:
We can use Main Operating Unit for all user.

kittiu: Hello, I think this may seem a valid point. If we don't fix it, it means, we can't setup Warehouses with no operating unit (which is also legit to me in some use case, where the whole company may want to allow user to cross warehouse).

But I also think that, this warning may come with good reason.

Can some experts in this area clarify this.

@max3903 @JordiBForgeFlow @HviorForgeFlow

kittiu: @Saran440 let's merge this as is for now? I think most people can get along with this (me included).

HviorForgeFlow: ping @AaronHForgeFlow aka OU master

AaronHForgeFlow: It seems this https://github.com/OCA/operating-unit/pull/234 is not included here. I think this would solve the issue. Can you check?

kittiu: Ok -> Done

stock_operating_unit/model/stock_location.py
```diff
@@ -0,0 +1,83 @@
...
+                    "|",
+                    "|",
+                    ("wh_input_stock_loc_id", "=", rec.ids[0]),
+                    ("lot_stock_id", "in", rec.ids),
+                    ("wh_output_stock_loc_id", "in", rec.ids),
+                ]
+            )
+            for w in warehouses:
+                if w.operating_unit_id and rec.operating_unit_id != w.operating_unit_id:
+                    raise UserError(
```
fmdl, line 28: use validationerror in check

PR #349, opened 2020-12-10
[13.0]add operating_unit_id to res.partner

luc-demeyer:


AaronHForgeFlow:
Could this be included in a separated module? Honestly, I prefer this than the module res_partner_operating_unit, which is a module I do not use but I think it creates some inconsistencies between users, partners and operating units related. I agree with the refactor/deactivation of that module but in any case I prefer this not to be included directly in the core module. Thanks.

luc-demeyer:
@AaronHForgeFlow 
I can move it to a seperate module but we first need to agree on the naming:
the logical name would be 'res_partner_operating_unit', hence we would first have to rename the current 'res_partner_operating_unit' to 'res_partner_multi_operating_unit'. 
I can make a PR for this and split out this PR into a seperate module once the rename is merged.

AaronHForgeFlow:
Ok for me. In any case the original author of that module should be aware of that @max3903 @bodedra

PR #351, opened 2020-12-16
[12.0][FIX] store computed field used in domain

kh-badep:
in a database with a lot of sale.order.line records, the ir.rule https://github.com/OCA/operating-unit/blob/2a4d17dab33feab02c8ab109b89e0e18348bac77/sale_operating_unit/security/sale_security.xml#L22 is applied on a computed non-stored field, which considerably slows down the navigation between Sale Order in form view.

This fix is intended to store the computed field so that the domain check will be only a select from a single table instead of a join from two.


GSLabIt (approved):
LGTM, tested with high number of sales order line

PR #254, opened 2020-02-13
[13.0][MIG] migration of module stock_account_operating_unit

alan196:
### Proposed changes

I've migrated the module stock_account_operating_unit and it's functionalities from Odoo 12.0 to 13.0. I've checked if the all unit tests are still running.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated necessary documentation

### Further comments

This module depends on the account_operating_unit  and stock_operating_unit module. That means that there should be a migration of it to 13.0. Because there is already a PR for migration I've added the changes of the PR to my local environment to test my migration.

This PR depends on the current migration PR:

https://github.com/OCA/operating-unit/pull/229
https://github.com/OCA/operating-unit/pull/231

PR #354, opened 2021-01-20
[13.0][MIG} stock account operating unit

robyf70:
Migration to v13


kittiu:
Duplicated with this? -> https://github.com/OCA/operating-unit/pull/254

Note: would be nice if you can change name of the PR to be stock_account_operating_unit, so it is easy to be found :)

robyf70:
@kittiu You are right! I'm closing mine. I would also add anglosaxon support to #254

PR #360, opened 2021-02-24
[13.0][FIX] hr_expense_operating_unit: multi operating-unit can't create expense sheet

Saran440:
Step to error:
1. Create expense from My Expenses
2. Change default operating unit.
3. Create Report
4. It throw error `Configuration error. The Operating Unit in the Expense sheet and in the Expense must be the same.`

Solve it by send operating unit from `hr.expense` to `hr.expense.sheet`


AaronHForgeFlow (approved):
Thanks for the fix and sorry for the late review :|

AaronHForgeFlow:
/ocabot merge patch

No need to forward port this, unable to reproduce in 14.0

PR #367, opened 2021-04-21
[13.0][FIX] hr_expense_operating_unit: default ou on move

Saran440:
Step to error:
1. Create expense from My Expenses with operating unit
2. Create Report > Submit to Manager > Approve > Post Journal Entries
Selection_005

3. Look at journal entry > operating unit on move is empty
Selection_006


Saran440:
Wrong commit move on https://github.com/OCA/operating-unit/pull/368

PR #368, opened 2021-04-21
[13.0][FIX] hr_expense_operating_unit: default ou on move

Saran440:
Step to error:

1. Create expense from My Expenses with operating unit
2. Create Report > Submit to Manager > Approve > Post Journal Entries
Selection_005

3. Look at journal entry > operating unit on move is empty
Selection_006

### DIFF
--- a/hr_expense_operating_unit/models/hr_expense.py
+++ b/hr_expense_operating_unit/models/hr_expense.py
@@ -67,11 +67,16 @@ class HrExpenseExpense(models.Model):
         return sheet
 
     def _get_account_move_line_values(self):
-        res = super(HrExpenseExpense, self)._get_account_move_line_values()
+        res = super()._get_account_move_line_values()
         for expense in self:
             res[expense.id][0].update({"operating_unit_id": self.operating_unit_id.id})
             res[expense.id][1].update({"operating_unit_id": self.operating_unit_id.id})
         return res
+
+    def _prepare_move_values(self):
+        move_values = super()._prepare_move_values()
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
The use of the operating unit in the account move is just for defaulting the operating unit in the account move lines when creating manual account_moves. 

I understand this may be confusing, perhaps we can explain that in the readme or even rename the field to be "Default Operating Unit" or similar. What do you think?

kittiu:
@AaronHForgeFlow you are right, it make sense, as the move line's OU is already passed in from expense.

And in theory, if there are many hr.expense(s) with different OU(s), the header is not relevant. So, in my opinion, this fix is not necessary.

But beside this, I just look in to the code and found this,
https://github.com/ecosoft-odoo/operating-unit/blob/13.0-fix-expense_default_ou_move/hr_expense_operating_unit/models/hr_expense.py#L69-L74

Shouldn't `self` should be `expense`
```
            res[expense.id][0].update({"operating_unit_id": expense.operating_unit_id.id})
            res[expense.id][1].update({"operating_unit_id": expense.operating_unit_id.id})

```

AaronHForgeFlow:
Hi @kittiu , I think you are right here, it should be expense.operating_unit_id.id 

Will you fix that?

kittiu:
> Hi @kittiu , I think you are right here, it should be expense.operating_unit_id.id
> 
> Will you fix that?

Ok -> https://github.com/OCA/operating-unit/pull/371

Saran440:
@AaronHForgeFlow @kittiu  Okay, I agree. Thank you.

PR #386, opened 2021-06-07
[14.0][ADD] analytic_operating_unit_access_all

kittiu:


AaronHForgeFlow (approved):
LGTM

Saran440:
@kittiu @AaronHForgeFlow  I found analytic ou has domain, so it will see user ou only.
I add new domain in analytic ou. please you check it again.

AaronHForgeFlow (approved):
Still good :)

Saran440:
Amm, I testing it still bug.

**Example**
My company has 5 OU
1. User A > there is access analytic all ou but User has 1 OU
2. In Analytic, User A still see 1 OU. (it should see 5 OU.)

kittiu:
> @kittiu @AaronHForgeFlow I found analytic ou has domain, so it will see user ou only.
> I add new domain in analytic ou. please you check it again.

@Saran440 I think what you try to fix is not the duty of this module (this module only want access to analytic from any document, and that only). If any, the fix should be in analytic_operating_unit (even that, I don't think it is needed).

Saran440:
Okay, I delete my commit already.

Saran440 (approved):

PR #402, opened 2021-07-12
[FIX] wrong code when adding OU to related partner

kh-badep:
res is super() result which will usually contains True. self must be used to compute the OU to add.

Current code gives the following error:
File "/OS/Users/Khalid/source/Odoo/OCA/operating-unit/res_partner_operating_unit/models/res_users.py", line 25, in write
    [(4, res.default_operating_unit_id.id)]
AttributeError: 'bool' object has no attribute 'default_operating_unit_id'


AaronHForgeFlow:
Closing, as already fixed in a different PR. Ty!

PR #397, opened 2021-06-23
[14.0][MIG] res_partner_operating_unit

Freni-OSI:

PR #403, opened 2021-07-13
[14.0][MIG] sale_multi_operating_unit

Vimal-OSI:
Dependency:

- [x] #401
- [ ] #397
- [x] #464


max3903 (approved):

Urvisha-OSI:
@max3903 we have added res_partner_operating_unit and sale_operating_unit modules in depends and those are not available in this branch that's why our test fails, you need to first merge res_partner_operating_unit PR and then after merging this one so it will not make any issue during the merge.

PR #412, opened 2021-08-17
[FIX] stock_operating_unit: fix warning in orderpoint constraint

alan196:
This PR fixes the following issue.

The constraint is checking a field that doesn't exists in stock.warehouse.orderpoint model.


AaronHForgeFlow:
/ocabot merge patch

fasttracking as clear bug

PR #413, opened 2021-08-17
[FIX] operating_unit: remove duplicate reference

sswapnesh:
Record `b2b_operating_unit` was referenced twice.

### DIFF
--- a/operating_unit/demo/operating_unit_demo.xml
+++ b/operating_unit/demo/operating_unit_demo.xml
@@ -13,7 +13,7 @@
     <record model="res.users" id="base.user_demo">
         <field name="default_operating_unit_id" ref="main_operating_unit" />
         <field
-            name="operating_unit_ids"
+            name="assigned_operating_unit_ids"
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):

AaronHForgeFlow (changes requested):
For some reason the operating units ids are empty for demo user in runbot. Can you check?

sswapnesh:
@AaronHForgeFlow  Updated, Not sure why it was not Working even we are assign it in 

https://github.com/OCA/operating-unit/blob/70a232bad1844cbe440c2a0ce8d0c37606973bbf/operating_unit/models/res_users.py#L64

flotho (changes requested):
Deleting some operating_units on the user don't seems to be stored properly : 
image

after validation  : 
image

sswapnesh:
@flotho I believe It has nothing to do with this PR. 
IMO this should be added as new Issue.

sswapnesh:
@AaronHForgeFlow  Can it be merged ?

AaronHForgeFlow (approved):
LGTM

AaronHForgeFlow:
@flotho I thing that is not related to this PR so that probably needs to be fixed in another PR. Do you agree?

PR #426, opened 2021-11-21
[14.0][ADD] account_asset_operating_unit, account_asset_operating_unit_access_all

ps-tubtim:
This PR adds 2 modules: account_asset_operating_unit, account_asset_operating_unit_access_all.

- **account_asset_operating_unit:** This module adds operating unit information to assets.
- **account_asset_operating_unit_access_all:** This module allow a user to have access to all OUs' asset, without having to add OUs in the user setting.


AaronHForgeFlow:
The depreciation journal entries should have the operating unit of the asset, shouldn't they?

ps-tubtim:
> The depreciation journal entries should have the operating unit of the asset, shouldn't they?

@AaronHForgeFlow I agree with you. Now, I have already updated this pull request to pass the operating unit to the depreciation journal entries. Could you review it? Thank you.

account_asset_operating_unit/models/account_asset_line.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import models
+
+
+class AccountAssetLine(models.Model):
+    _inherit = "account.asset.line"
+
+    def _setup_move_data(self, depreciation_date):
+        move_data = super()._setup_move_data(depreciation_date)
+        move_data.update({"operating_unit_id": self.asset_id.operating_unit_id.id})
```
AaronHForgeFlow:
```suggestion
        if self.asset_id.operating_unit_id
            move_data.update({"operating_unit_id": self.asset_id.operating_unit_id.id})
```

AaronHForgeFlow (commented):
Code review. LG only one comment. A new test that checks the OU in the asset would be fine too.

account_asset_operating_unit/models/account_asset_line.py
ps-tubtim: I have already updated it. Thank you.

Saran440 (approved):
Functional Test :+1:

AaronHForgeFlow (approved):

PR #428, opened 2021-12-02
[14.0][FIX]purchase_operating_unit: Operating Unit Issue when creating RFQ in Purchase App

Vimal-OSI:
If you have a default OU set on your user get the following error when creating a RFQ "No Warehouse found with the Operating Unit indicated in the Purchase Order"

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,7 @@ repos:
           - --remove-duplicate-keys
           - --remove-unused-variables
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
   - repo: https://github.com/pre-commit/mirrors-prettier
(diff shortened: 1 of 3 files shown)


kittiu:
@Vimal-OSI just curious. What is the benefit of this fix, except you are using onchange instead of UserError?

Just curious, I might not technically enough to understand :)

Vimal-OSI:
> @Vimal-OSI just curious. What is the benefit of this fix, except you are using onchange instead of UserError?
> 
> Just curious, I might not technically enough to understand :)

Well!, If we have used "raise UserError" on On change method then It's blocks process if condition is satisfied. (I believe, It's not good practice to raise warning on on-change method).
instead of it, we can use return warning, then won't be block process but raised alert to user too.

AaronHForgeFlow:
If the user tries to confirm the PO and he gets an error then this is fine to me. And that is what it is happening I believe

AaronHForgeFlow:
Hi @Vimal-OSI, can you please explain the exact steps to reproduce the issue? I tried the following: assign myself the operating unit B2B as default. Then create an RFQ and the default picking type for B2B receipts was selected, then I tried to modify the picking type field but I got no errors no warning messages.

max3903 (approved):

PR #429, opened 2021-12-06
[14.0][FIX] hr_expense_operating_unit: filter domain OU on expense

Saran440:
On expense, User can select `Analytic Account` difference OU.

Fix `Analytic Account` can select following operating unit or Analytic Account is not OU.
Selection_004


Saran440:
Closed, We can do it when install module hr_expense_operating_unit and analytic_operating_unit only.

PR #434, opened 2021-12-16
[14.0][FIX] hr_expense_operating_unit: check OU on expense line difference

Saran440:
[BUG] 
Can not `post journal entries` if OU are difference

Step to error

1. Create 2 expenses (OU1, OU2)
2. Create 1 report from 2 expenses
4. Delete Operating Unit in expense report (header)
4. Change Operating Unit in expense
Selection_003
5. Save -> it should can't save
6. Submit to manager > Approve > Post Journal Entries
7. it's will error singleton
Selection_004

[FIX]
- Add constrains on expense report. check ou on expense line must be the same.
- Fix update OU on move line from expense instead of self


ps-tubtim (approved):
Code LGTM.

PR #436, opened 2021-12-26
[14.0][FIX] Access all OU on stock.move

newtratip:
If OU was defined in destination location which difference with user's OU, it will throw access error
Selection_803

This PR is fixed.

cc: @kittiu


AaronHForgeFlow (approved):

Saran440 (approved):
LGTM :+1:

PR #440, opened 2022-01-09
[14.0] stock_operating_unit: Add Record Rule to Quants.

gdgellatly:
Without this rule, a user will not be able to view the inventory report if they do not belong to all operating units of all the companies they belong.

Performance may be a cost here, in future versions we may want to store operating_unit_id on the quants directly as a relational stored field.


AaronHForgeFlow (approved):
LGTM

alan196 (approved):

PR #455, opened 2022-02-01
[10.0] sale_stock_operating_unit patch: omitted super() in onchange_operatin…

whulshof:
…g_unit_id(), so the on change method in sale_operating_unit module did not get called.


AaronHForgeFlow (approved):
Code review LGTM

CI is failing because an error in the module sale_crm_operating_unit, I guess something changed in the core

PR #456, opened 2022-02-08
[14.0][FIX] account_operating_unit: Accounting Dashboard Payments #361

Vimal-OSI:
Fixes for #361 
Reproducible steps (with EE Addons):
1. login with admin or any accounting user.
2. Go to accounting -> Dashboard -> Select any bank type journal and click on three dots from right top corner.
3. Select Internal Transfer option, it will open new account payment record
4. Keep all the default values as it is and fill amount and memo field and try to save.

Error:
It will throw an error saying the account.move(7) is not exist or deleted.
Screenshot from 2022-02-08 14-54-54

PR #463, opened 2022-03-09
[14.0][FIX] account_operating_unit calculate proper OU in payments regardless of their source

kh-badep:
This is an alternative approach to: https://github.com/OCA/operating-unit/pull/456.


AaronHForgeFlow (approved):
I agree this solution is more secure

d2isolutions:
Hi @AaronHForgeFlow this fix solved an issue I was having. I see you approved these changes. Is there a reason why it was never merged?

d2isolutions:
@AaronHForgeFlow disregard my previous comment I see that PR 456 solution is preferred over this one.

PR #464, opened 2022-03-09
[14.0][MIG] sale_operating_unit

olveracuenca:


alan196 (approved):
LGTM 👍

AaronHForgeFlow:
It seems test fail because needed permission in stock.move.line, that is weird because this module does not depend on sale_stock :confused:

alan196:
@olveracuenca 

The module `stock_operating_unit` is already migrated and for that reason, the `stock` module will be installed and module `sale_stock` is an auto_install module.

You need to check the test and validate the product you are using, you can modify the test with a product that is a service to avoid this issue.

AaronHForgeFlow:
Hi @olveracuenca , will you continue with this PR? I think the only needed thing is to create a service product in the test instead of a storable product. WDYT?

alan196 (approved):
LGTM 👍

olveracuenca:
hi @AaronHForgeFlow , yes  changing the product to 'service' type product it runs correctly.

olveracuenca:
@AaronHForgeFlow  Thanks.

AaronHForgeFlow (commented):
For some reason the onchange_team_id method is not working. I tried to change the team in a sales quotation and the operating unit is not taken from the sales team, and moreover, I cannot select the operating unit from the team, only Main OU shows:
Peek 2022-04-22 07-53

gdgellatly:
@AaronHForgeFlow yes it is an annoying behaviour, before operating unit set sales team, now it occurs other way around, so I found you need to have your salesperson in the sales team of the operating unit.

AaronHForgeFlow:
@gdgellatly, you are correct, thank you. I think we can go forward if this is mentioned in the roadmap.

sale_operating_unit/views/sale_view.xml
```diff
@@ -0,0 +1,70 @@
...
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <field name="user_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    widget="selection"
+                    domain="[('user_ids', 'in', uid)]"
+                    groups="operating_unit.group_multi_operating_unit"
+                />
+            </field>
```
Saran440, line 46: I'm not sure that sale module should be add filter domain analytic following OU.
```
<xpath expr="//field[@name='analytic_account_id']" position="attributes">
    <attribute name="domain">['|', ('company_id', '=', company_id), ('company_id', '=', False), '|', ('operating_unit_ids', '=', operating_unit_id), ('operating_unit_ids', '=', False)]</attribute>
</xpath>
```

Saran440 (commented):
Functional Test. Okay
I'm not sure that sale module should be add filter domain analytic following OU.

WDYT?

gdgellatly:
> @gdgellatly, you are correct, thank you. I think we can go forward if this is mentioned in the roadmap.

@AaronHForgeFlow Interestingly I got a bug report today that may mean we have to resolve this issue. The scenario is if a person belongs to multiple operating units, but duplicates a sales order for which he is not a favorite member of that sales team/operating unit. I haven't had a chance to fully check yet, but worth flagging now.

francesco-ooops:
Hi @olveracuenca , any plans to finish this one?

@AaronHForgeFlow and @Saran440 have your comments been addressed with latest commits?

max3903 (approved):

PR #468, opened 2022-04-12
[14.0][FIX] account_operating_unit: fix issues shows error when creating an unlinked internal transfer

olveracuenca:


AaronHForgeFlow:
Can you please describe the steps to reproduce the error? Thanks in advance

AaronHForgeFlow:
Can you please describe the steps to reproduce the error? Thanks in advance

dominiquekonsuntack:
open the accounting App. Go to a bank journal, and select "internal transfer". Save the record. You will have an error message:
New_-_Odoo

The error is that in the when a payment in registered independently from an invoice, the active_ids is the id of the account.payment - there is no invoice ('account.move') as origin document.
odoo14_–_account_operating_unit_…_account_payment_py

I don't think the suggested fix makes any sense...
I suggest instead another fix:

    def _prepare_move_line_default_vals(self, write_off_line_vals=None):
        res = super()._prepare_move_line_default_vals(write_off_line_vals)
        for line in res:
            line["operating_unit_id"] = self.operating_unit_id.id

        # invoices = self.env["account.move"].browse(self._context.get("active_ids"))          # removed

        ids = self._context.get('active_ids')                                                 # added
        invoices = self.env['account.move'].search([('id', 'in', ids)])         # added

        invoices_ou = invoices.operating_unit_id
        if invoices and len(invoices_ou) == 1 and invoices_ou != self.operating_unit_id:
            destination_account_id = self.destination_account_id.id
            for line in res:
                if line["account_id"] == destination_account_id:
                    line["operating_unit_id"] = invoices_ou.id
        return res

olveracuenca:
hi @dominiquekonsuntack 
thanks for your correction and update the lines and try

account_operating_unit/models/account_payment.py
```diff
@@ -25,7 +25,8 @@ def _prepare_move_line_default_vals(self, write_off_line_vals=None):
         res = super()._prepare_move_line_default_vals(write_off_line_vals)
         for line in res:
             line["operating_unit_id"] = self.operating_unit_id.id
-        invoices = self.env["account.move"].browse(self._context.get("active_ids"))
+        ids = self._context.get("active_ids")
+        invoices = self.env["account.move"].search([("id", "in", ids)])
```
AaronHForgeFlow: but if  active_ids are payments then the searching is incorrect isn't it? I think you should check the active_model and if the active_model is "account.payment" do another search

olveracuenca: hi @AaronHForgeFlow 
thanks, let me analyze the code again

AaronHForgeFlow (commented):
Code LGTM.

Why did you downgrade the module to alpha development_status? In any case that has to be done in the manifest.

alan196 (approved):
👍  Functional and technical review.

alan196:
@AaronHForgeFlow code is now only changin python

PR #471, opened 2022-04-19
[14.0][IMP] travis to github-actions

AaronHForgeFlow:
@ForgeFlow


AaronHForgeFlow:
@sbidoul @gurneyalex  sorry for ping you here. But I cannot make github actions to run tests on this repo. Do you know what I have to do? Thanks in advance.

sbidoul:
@AaronHForgeFlow `.test copy.yml` does not look like the correct file name. Also 1.5.2 is not the latest version of the template.
Did you run `copier update` to update the config ?

AaronHForgeFlow:
Thank you @sbidoul I was doing some tests, and I uploaded the wrong file. Also the copier update was failing before. Now it works. Thanks for the help :)

AaronHForgeFlow:
@sbidoul @kittiu  FYI: I am fixing the license on the mis-builder modules on this repo, they should be AGPL.

@kittiu FYI I am adding a development status for account_asset_operating_unit_access_all. It is the same as its dependency

AaronHForgeFlow:
fail

MiquelRForgeFlow (approved):

PR #495, opened 2022-08-08
[ADD][14.0] payroll_operating_unit: add OU to a payslip based on the contract

mtelahun:
This module introduces an operating unit on a payslip. By default the OU on the contract is used, if available. However, the user may also set it manually.


AaronHForgeFlow:
Have you checked https://github.com/OCA/operating-unit/tree/14.0/hr_payroll_account_operating_unit?

mtelahun:
Hi @AaronHForgeFlow . I see that this PR is redundant then. I'll close it. Thanks for letting me know :grin:

PR #499, opened 2022-08-16
[14.0][FIX] account_asset_operating_unit: fix reporting by OU

ps-tubtim:
This PR fixes reporting when generating asset reports with the asset group and operating unit.

### DIFF
--- a/account_asset_operating_unit/report/account_asset_report_xls.py
+++ b/account_asset_operating_unit/report/account_asset_report_xls.py
@@ -28,20 +28,15 @@ class AssetReportXlsx(models.AbstractModel):
     def _get_assets(self, wiz, data):
         super()._get_assets(wiz, data)
         if wiz.operating_unit_id:
-            # Filter asset by operating unit
             assets = data["assets"]
             assets_by_ou = assets.filtered(
-                lambda a: not a.operating_unit_id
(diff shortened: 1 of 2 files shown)


account_asset_operating_unit/report/account_asset_report_xls.py
```diff
@@ -28,20 +28,16 @@ def _get_asset_template(self):
     def _get_assets(self, wiz, data):
         super()._get_assets(wiz, data)
         if wiz.operating_unit_id:
-            # Filter asset by operating unit
             assets = data["assets"]
             assets_by_ou = assets.filtered(
                 lambda a: not a.operating_unit_id
                 or a.operating_unit_id == wiz.operating_unit_id
```
newtratip: I think if filter by operating unit should show only assets that match with selected opearting unit on the wizard. Should not show assets that no have operating unit.

this line, if I change to
```
assets_by_ou = assets.filtered(lambda a: a.operating_unit_id == wiz.operating_unit_id)
```

Is this ok for you ? @ps-tubtim

ps-tubtim: @newtratip I agree with you. I already change it.

Saran440 (approved):

Nantikan23 (approved):
Functional Tested

PR #497, opened 2022-08-08
[15.0][ADD] pos_operating_unit: new module to manage POS with operating unit

alan196:
### Proposed changes

New module to manage Operating Unit in PO

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

### Comments

This module depends on 
https://github.com/OCA/operating-unit/pull/478
https://github.com/OCA/operating-unit/pull/420

PR #502, opened 2022-09-01
[14.0][ADD] pos_operating_unit_access

ilyasProgrammer:
Access to POS related models, based on operating units.


francesco-ooops (approved):
functional review ok!

---

As per description + roadmap:

This module adds a field "operating unit" in POS with the only goal of restricting visibility (therefore access) to users with same OU allowed.

This also applies to sessions, orders and payments in POS.

A "Point of sale" field is added to orders and payments for clarity.

This module only manages visibility of POS and related orders. If a more complete
"pos_operating_unit" is proposed and merged, then it should cover features of this
module which can be discontinued.

@OCA/accounting-maintainers what do you think? 

ping also to @alan196 since he's been working on pos_operating_unit in #497

SicurSam (approved):
Functional review ok!

ilyasProgrammer:
https://github.com/OCA/operating-unit/pull/521

PR #510, opened 2022-10-06
[13.0][FIX] sale_operating_unit: fix perfomance issues in sale order

alan196:
This PR makes the field operating_unit_id stored in sale order line model. 

This is required because it causes performance issues in databases with large number of sale.order records.

Without this change, if you open a sale order record with a database with 1 million records, it takes 20 seconds to open the sale.order.

With this change, it takes less than 1 second.

### DIFF
--- a/sale_operating_unit/models/sale_order.py
+++ b/sale_operating_unit/models/sale_order.py
@@ -73,5 +73,5 @@ class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
     operating_unit_id = fields.Many2one(
-        related="order_id.operating_unit_id", string="Operating Unit"
+        related="order_id.operating_unit_id", string="Operating Unit", store=True,
     )


alan196:
@LoisRForgeFlow

AaronHForgeFlow:
Ok for me. Can you see why travis is failing? and also, is it possible to add a migration script so the column is filled more quickly?

alan196:
@nhomar @moylop260 Could you help me to review this PR 🙏

moylop260 (approved):
👍

AaronHForgeFlow:
Hi @alan196 , when you apply this change to a database. Does it fill the column for the already created sale order lines? or do they remain blank? In case they are filled in does it take to much to fill that value?  I just wonder if the migration scripts is really needed or not.

moylop260:
Hi @AaronHForgeFlow 

Good catch!

@alan196 
Please, create a migration script since that if it is slow opening the view so it will be slow computing the whole database for old databases using this modules and using this new change

alan196:
@AaronHForgeFlow I apply this change in a database with millions of sale order lines and the migration script was not necessary and it was fast

AaronHForgeFlow (approved):
Ok, if the columns are filled in when updating the module and it is not slow we can merge it. Can you do the forward ports to next versions?

PR #513, opened 2022-10-25
[14.0][UPT] template from copier

### LINKED ISSUE #505: repo readme.md
Hi, would it be possible to fix the readme.md for this repo to align it to the other repos? this is how 14 appears:

image

I also see in v12 and 13 there are still links to transifex

AaronHForgeFlow:
Fixes https://github.com/OCA/operating-unit/issues/505

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
 lib64/
 parts/
 sdist/
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,3 +1,5 @@
+
+
 [MASTER]
 load-plugins=pylint_odoo
 score=n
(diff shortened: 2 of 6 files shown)


.pylintrc
```diff
@@ -1,3 +1,5 @@
+
+
```
MiquelRForgeFlow, line 2: wrong

.gitignore
```diff
@@ -15,7 +15,6 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
```
MiquelRForgeFlow, line 18: wrong

.pylintrc
AaronHForgeFlow: why? I just did copier update, it should work like this, isn't it?

MiquelRForgeFlow: yes, but it's ugly having double space

AaronHForgeFlow: that's bad

Saran440:
@AaronHForgeFlow Hello, This PR conflict, do you still do this PR or will move it to https://github.com/OCA/operating-unit/pull/518

AaronHForgeFlow:
Hi @Saran440  We can follow up in your PR

PR #518, opened 2022-11-21
[14.0][FIX] update copier template

Saran440:


.pylintrc
```diff
@@ -1,3 +1,5 @@
+
+
```
MiquelRForgeFlow: avoid white spaces

.pylintrc-mandatory
```diff
@@ -1,3 +1,4 @@
+
```
MiquelRForgeFlow:
```suggestion
```

MiquelRForgeFlow (approved):

Saran440:
@MiquelRForgeFlow Thanks for review, i fix it BUT other repo is space too. may be it unnecessary.

AaronHForgeFlow (approved):

PR #521, opened 2022-11-22
[14.0][ADD] pos_operating_unit

ilyasProgrammer:
replacement for https://github.com/OCA/operating-unit/pull/502


francesco-ooops (approved):
functional review ok!

also added test coverage

---

As per description + roadmap:

This module adds a field "operating unit" in POS with the only goal of restricting visibility (therefore access) to users with same OU allowed.

This also applies to sessions, orders and payments in POS.

A "Point of sale" field is added to orders and payments for clarity.

This module only manages visibility of POS and related orders. If a more complete
"pos_operating_unit" is proposed and merged, then it should cover features of this
module which can be discontinued.

@OCA/accounting-maintainers good for merge?

SicurSam (approved):
functional review ok!

francesco-ooops:
@OCA/accounting-maintainers gentle reminder :)

francesco-ooops:
@OCA/accounting-maintainers when you can, thanks

francesco-ooops:
@OCA/accounting-maintainers gentle reminder!

francesco-ooops:
@pedrobaeza could you take a look? thank you!

pedrobaeza:
Sorry, not using operating-unit nor PoS.

AaronHForgeFlow (commented):
Sorry for being picky, but I think the module name should be pos_operating_unit.

Also, the module was proposed in v15 here: https://github.com/OCA/operating-unit/pull/497 by @alan196 It was proposed before this one. Theoretically this PR should take over, because it was created first. In case you have a different opinion, I will let you to discuss what version of the module should be in OCA at the end. But please notice that only one version should be in OCA, and the module should work the same in all versions.

francesco-ooops:
> Sorry for being picky, but I think the module name should be pos_operating_unit.
> 
> Also, the module was proposed in v15 here: #497 by @alan196 It was proposed before this one. Theoretically this PR should take over, because it was created first. In case you have a different opinion, I will let you to discuss what version of the module should be in OCA at the end. But please notice that only one version should be in OCA, and the module should work the same in all versions.

Thanks for your feedback @AaronHForgeFlow !

We have no issues renaming the module, although this is not a full "operating unit" implementation for POS, but rather a restriction of access for users according to their operating unit - as described in "roadmap".

Since in general there isn't much activity in this repo and PR #497 was getting stale, we thought of making a different module than @alan196 's PR as the scopes are different. 

If that's ok for you, we can rename the module as pos_operating_unit, make the scope clear in description and then it can be expanded in further PRs.

AaronHForgeFlow:
Well, looking at the code it seems this implements access control and also operating unit consistency across records. That is what most modules do in this repo. So for me it makes more sense to call it just pos_operating_unit, but in any case, I am not using POS and I am not familiar with it so I will let others to give their opinions. My only goal here is to avoid duplicated efforts and overlapped functionality here.

francesco-ooops:
Thanks, it doesn't look like there's much interest on the matter and the other PR seems far from active, so we'll just rename the module and ask for merge again.

francesco-ooops:
@AaronHForgeFlow renamed, do you think we can merge now? :)

francesco-ooops:
@AaronHForgeFlow gentle reminder

AaronHForgeFlow:
@francesco-ooops I am fine with it. Just wondering if @alan196 agrees.

alan196 (approved):
👍 I'm okay with this

PR #524, opened 2023-01-05
[14.0][FIX] OperatingUnit.name_search() omits args domain in "other search"

MarcinSokolowski-Solvti:
Fix: omit domain passed in args to OperatingUnit.name_search() in "other search"

Currently when we create model with relation to operating.unit with domain (for example Many2one field with ('company_id','=',company_id) domain) and make field available in form view, users will be able to select only operating units that fulfill domain condition, unless they start to type name to field to search for it. Then all operating units that user have access to will be available, not taking into account domain condition. 

This PR fixes that issue.

### DIFF
--- a/operating_unit/models/operating_unit.py
+++ b/operating_unit/models/operating_unit.py
@@ -51,6 +51,8 @@ class OperatingUnit(models.Model):
         names2 = []
         if name:
             domain = [("code", "=ilike", name + "%")]
+            if args:
+                domain.extend(list(args))
             names2 = self.search(domain, limit=limit).name_get()
         # Merge both results
         return list(set(names1) | set(names2))[:limit]


operating_unit/models/operating_unit.py
```diff
@@ -51,6 +51,8 @@ def name_search(self, name="", args=None, operator="ilike", limit=100):
         names2 = []
         if name:
             domain = [("code", "=ilike", name + "%")]
+            if args:
+                domain.extend(list(args))
```
AaronHForgeFlow, line 55: Would be possible a case when args can be not transformed into a list?

AaronHForgeFlow (commented):
Makes sense, just a question regarding the implementation

operating_unit/models/operating_unit.py
MarcinSokolowski-Solvti: I don't think so, even in default Odoo implementation of `name_search` `args` are treated as transformable to list or Falsy only

AaronHForgeFlow: I see. Can you please change the commit message so it begins with `[FIX] operating_unit: ` ? then I think it is fine

MarcinSokolowski-Solvti: Amended commit message

StefanWiselka-Solvti: Would be nice to have this fix in branches 15.0, 16.0 and 13.0

StefanWiselka-Solvti (approved):

AaronHForgeFlow (approved):
LGTM! I will just leave it open for a couple of days before merging.

AaronHForgeFlow:
/ocabot merge patch

@MSokol00 can you do the forward port?

PR #534, opened 2023-01-18
[15.0][FIX]-account_operating_unit: fix the error 'You are trying to …

hieulucky111:
[15.0][FIX]-account_operating_unit: fix the error 'You are trying to reconcile some entries that are already reconciled'

"You are trying to reconcile some entries that are already reconciled" 
How to reproduce the issue:
1. Use Payment Term on Invoice, so Payable / Receivable Journal Item Line will be split to 2 lines.
2. Register Payment to pay the first Payable / Receivable Journal Item Line.
3. Second Register Payment will face the error above.

image


AaronHForgeFlow:
@hieulucky111  why did you close this? is it not an issue anymore?

Edit:  I see. There is a supersede here https://github.com/OCA/operating-unit/pull/535

PR #400, opened 2021-06-29
[12.0][FIX] analytic_operating_unit: Error search OUs in analytic account

Borruso:

PR #539, opened 2023-01-24
[15.0][FIX] analytic operating unit: domain ou in users

Saran440:
[BUG]
user (multi ou) can't see other ou in analytic

This pr fix domain in analytic can select other ou in analytic

### DIFF
--- a/analytic_operating_unit/models/account_analytic_account.py
+++ b/analytic_operating_unit/models/account_analytic_account.py
@@ -11,7 +11,6 @@ class AccountAnalyticAccount(models.Model):
         comodel_name="operating.unit",
         string="Operating Units",
         relation="analytic_account_operating_unit_rel",
-        domain="[('user_ids', '=', uid)]",
         column1="analytic_account_id",
         column2="operating_unit_id",
     )


AaronHForgeFlow (approved):
Thanks for the fix. Can you put the module name in the commit message? just for better reading when looking at the history.

MiquelRForgeFlow (changes requested):
Incomplete fix. The complete one is https://github.com/OCA/operating-unit/pull/400. Could you cherry-pick it instead or complete your own? Also a Pr for v13 and v14 should be needed, right?

Saran440:
@MiquelRForgeFlow I tested with version15 it's not working.
and why we have this domain? I think user can see ou following access right.

Peek 2023-01-26 18-22

analytic_operating_unit/views/analytic_account_view.xml
```diff
@@ -11,6 +11,7 @@
                 <field
                     name="operating_unit_ids"
                     widget="many2many_tags"
+                    domain="[('user_ids', '=', uid)]"
```
MiquelRForgeFlow: Oh well, if the module doesn't work with this line, then remove it :S I just thought that it was needed as it was solved this way in v12.

AaronHForgeFlow: I confirm that fix do not work in v15. Can you remove that part?

Saran440:
@MiquelRForgeFlow @AaronHForgeFlow I fixed it by cherry-pick https://github.com/OCA/operating-unit/pull/400 and remove domain in xml file

MiquelRForgeFlow (approved):

PR #540, opened 2023-01-27
[FIX] 14.0 sale_operating_unit Error sale report to add fiel operating unit

cramosh:
This PR solves the error that the sale report has when adding the sale_operating_unit field after the odoo update


AaronHForgeFlow:
Thank you for the fix, just a couple of things:
* You can add yourself as contributor in the readme, but the Copyright section on top is reserved by the original authors on the module only
* Can you please add the module name in the commit message?
* Also please include the module name in the title of this PR

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

PR #555, opened 2023-02-01
[15.0][MIG] account_asset_operating_unit

Saran440:
- Add test script
- Add group by operating unit


AaronHForgeFlow (approved):
Code review

ps-tubtim (approved):

Saran440:
@AaronHForgeFlow Hello, May I ask you about filter OU.

Asset1 - OU-A
Asset2 - No OU

Options
1. Filter OU-A --> it will generate all OU-A and asset don't have operating unit (Asset1 and Asset2)
2. Filter OU-A --> it will generate all OU-A only (Asset1). - https://github.com/OCA/operating-unit/pull/499

What is option that make sense for you?

AaronHForgeFlow:
Typically, for access meaning if I have access to OU-A then I would see all assets OU-A and assets without OU.

However, for the purpose of filtering I think it makes more sense the second option. So yes, probably that PR needs to be forward ported.

PR #559, opened 2023-02-02
Fix: [14.0] Ensure active_model is account.move before attempting to browse

dannyadair:


gdgellatly (approved):
tested with prod data. stops errors when assignation comes from payment orders.

Saran440 (approved):
Code Reviewed. :+1: 

Can you squash commit into 1 please?

account_operating_unit/models/account_payment.py
```diff
@@ -25,11 +25,16 @@ def _prepare_move_line_default_vals(self, write_off_line_vals=None):
...
         for line in res:
             line["operating_unit_id"] = self.operating_unit_id.id
-        invoices = self.env["account.move"].browse(self._context.get("active_ids"))
-        invoices_ou = invoices.operating_unit_id
-        if invoices and len(invoices_ou) == 1 and invoices_ou != self.operating_unit_id:
-            destination_account_id = self.destination_account_id.id
-            for line in res:
-                if line["account_id"] == destination_account_id:
-                    line["operating_unit_id"] = invoices_ou.id
+        if self.env.context.get("active_model") == "account.move":
```
AaronHForgeFlow, line 28: In order to reduce the diff I prefer to put something like

```
 if not self.env.context.get("active_model", False) == "account.move":
    return res
```

But in any case it is more important to fix the bug. Can you squash commits and add the module name in the commit message? Thank you!

AaronHForgeFlow:
Superseded here to fix commit messages: https://github.com/OCA/operating-unit/pull/570

PR #562, opened 2023-02-14
[15.0][ENH] hr_operating_unit: add OU to department

ps-tubtim:
Adds an operating unit to the department

Selection_182

### DIFF
--- a/hr_operating_unit/views/hr_department_views.xml
+++ b/hr_operating_unit/views/hr_department_views.xml
@@ -8,7 +8,6 @@
                 <field
                     name="operating_unit_id"
                     options="{'no_create': True}"
-                    domain="[('company_id','=', company_id)]"
                     groups="operating_unit.group_multi_operating_unit"
                 />
             </field>
--- a/hr_operating_unit/views/hr_department_views.xml
+++ b/hr_operating_unit/views/hr_department_views.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_department_form" model="ir.ui.view">
+        <field name="model">hr.department</field>
+        <field name="inherit_id" ref="hr.view_department_form" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    options="{'no_create': True}"
+                    domain="[('company_id','=', company_id)]"
+                    groups="operating_unit.group_multi_operating_unit"
+                />
+            </field>
+        </field>
+    </record>
+    <record id="view_department_tree" model="ir.ui.view">
+        <field name="model">hr.department</field>
+        <field name="inherit_id" ref="hr.view_department_tree" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="after">
+                <field
+                    name="operating_unit_id"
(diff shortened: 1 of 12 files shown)


hr_operating_unit/views/hr_department_views.xml
```diff
@@ -0,0 +1,49 @@
...
+    <record id="view_department_form" model="ir.ui.view">
+        <field name="model">hr.department</field>
+        <field name="inherit_id" ref="hr.view_department_form" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    options="{'no_create': True}"
+                    domain="[('company_id','=', company_id),
+                       ('user_ids', 'in', uid)]"
```
Saran440:
```suggestion
                       ]"
```

hr_operating_unit/views/hr_department_views.xml
```diff
@@ -0,0 +1,49 @@
...
+        </field>
+    </record>
+    <record id="view_department_tree" model="ir.ui.view">
+        <field name="model">hr.department</field>
+        <field name="inherit_id" ref="hr.view_department_tree" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="after">
+                <field
+                    name="operating_unit_id"
+                    domain="[('user_ids', 'in', uid)]"
```
Saran440:
```suggestion
                    
```

Saran440 (commented):
Please, delete filter with domain users. it will see 1 ou only if you add domain user.

hr_operating_unit/views/hr_department_views.xml
ps-tubtim: @Saran440 I already updated it.

hr_operating_unit/views/hr_department_views.xml
ps-tubtim: @Saran440 I already updated it.

Saran440 (approved):
LGTM :+1:

Saran440:
This PR fixed domain users in hr.employee

Selection_012

TheerayutEncoder (approved):
LGTM :100:

AaronHForgeFlow:
I am not a user of this module, but this change seems legit. I would recommend to do a migration script to prefill the information in the department.

Saran440:
@AaronHForgeFlow I think this enhance shouldn't migration script because all department will default 1 OU

Example:

Case 1 (No script)
Deparmtent A - No OU
Deparmtent B - No OU
Deparmtent C - No OU
Deparmtent D - No OU

Case 2 (there is script)
Deparmtent A - OU 1
Deparmtent B - OU 1
Deparmtent C - OU 1
Deparmtent D - OU 1

Case1 all user can see department
Case2 user OU1 can see all department, it must change some department to correct OU
I think case1 is makesense more than

What do you think?

AaronHForgeFlow:
@Saran440 , I think that makes sense.

Would it make sense to put a restriction so the operating unit of the employee cannot be different from the operating unit of the department?

In any case I am not using this module and I prefer actual users to discuss this.

AaronHForgeFlow (commented):
@ps-tubtim  as the change is quite major and I am not using this module I would rather like to have more approvals from actual users.  Code LG

PR #564, opened 2023-02-17
[15.0][ADD] account_asset_transfer_operating_unit

Saran440:
This module add operating unit in asset transfer

At header in asset wizard will default OU following origin asset (if there is 1 ou)
At `To new asset` tab, default following header

If you need create move with multi operating unit, not select OU in header (select in lines only)

image

Depend on:
- [x] account_asset_transfer - https://github.com/OCA/account-financial-tools/pull/1507
- [x] account_asset_operating_unit - https://github.com/OCA/operating-unit/pull/555


Nantikan23 (approved):
Functional Tested

AaronHForgeFlow (approved):
Code review LG

PR #423, opened 2021-10-31
[14.0][FIX] account_operating_unit: hook function and check inter ou on company

Saran440:
move on function add moveline on payment to new function
and check inter ou on company, if not it can overwrite line without change stage.

PR #565, opened 2023-02-28
[15.0][FIX] account_operating_unit: re reconcile payment for case self balance only

Saran440:
Reference : https://github.com/OCA/operating-unit/pull/423/files

This pr is fixed for not re reconcile if payment is not self balance
because may be other module will have a issue if payment is set to draft


AaronHForgeFlow (approved):
Code review LG!

ps-tubtim (approved):
LGTM.

AaronHForgeFlow:
Can you include the module name in the commit message? sorry for the nitpicking

Saran440:
@AaronHForgeFlow Fixed :)

PR #573, opened 2023-03-14
[15.0][FIX] account_operating_unit: add OU to tree view of account.move.line

ps-tubtim:
I add OU to the tree view of `account.move.line`.

`view_move_line_tree_grouped` has higher priority than `view_move_line_tree`.
So, when you want to show a tree view of `account.move.line` by default is used `view_move_line_tree_grouped`.


AaronHForgeFlow (approved):
Agree

Saran440 (approved):
LGTM :+1:

PR #576, opened 2023-03-17
[16.0][MIG] operating_unit: migration to 16.0

huguesdk:


alan196 (approved):
👍 Technical review

AaronHForgeFlow (commented):
Functional review. I can remove operating unit for users except for the admin, If I remove the operating unit for the admin and I jump out and jump in again the user form view of the admin I see all OUs are assigned. Can't remove. Could you check that?

houssine78:
Hi @huguesdk, did you see the comment from Aaron ? Does it needs action from your side ? It's an improvement or this should work as the Aaron's expectations ?

Thank you :)

huguesdk:
hi @houssine78. yes, i saw the comment, but thanks for the heads up! i did not intend to change the behavior, so if it does not work as in 15.0, then i need to fix it.

@AaronHForgeFlow: thanks for the review! do you know if this is something that works as expected in 15.0?

operating_unit/models/operating_unit.py
```diff
@@ -41,22 +42,6 @@ class OperatingUnit(models.Model):
         ),
     ]
 
-    @api.model
-    def name_search(self, name="", args=None, operator="ilike", limit=100):
```
gurneyalex: is there a reason for the removal of the name_search method?

gurneyalex (approved):

operating_unit/models/operating_unit.py
huguesdk: yes, it is because the new `_rec_names_search` class variable now allows for the same behavior without the need to define `.name_search()`. see [the guide to migrate to 16.0](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#tasks-to-do-in-the-migration).

huguesdk:
> Functional review. I can remove operating unit for users except for the admin, If I remove the operating unit for the admin and I jump out and jump in again the user form view of the admin I see all OUs are assigned. Can't remove. Could you check that?

@AaronHForgeFlow: i tested in 15.0 and the behavior is the same. it comes from [this code](https://github.com/OCA/operating-unit/blob/15.0/operating_unit/models/res_users.py#L74), which forces the users belonging to the `operating_unit.group_manager_operating_unit` group to have all operating units linked to them.

AaronHForgeFlow:
Thanks for checking @huguesdk I think that is ok then. I just won't assign users to that group when they don't have access to all OU.

AaronHForgeFlow (approved):

PR #577, opened 2023-03-17
[15.0][MIG] project_operating_unit: migration to 15.0

huguesdk:


project_operating_unit/models/project_task.py
```diff
@@ -11,7 +11,4 @@ class ProjectTask(models.Model):
         comodel_name="operating.unit",
         related="project_id.operating_unit_id",
         string="Operating Unit",
-        default=lambda self: self.env["res.users"].operating_unit_default_get(
```
alan196: Why do you remove the default?

alan196 (changes requested):
Only one comment

project_operating_unit/models/project_task.py
huguesdk: thanks for the review!

with the default, odoo gives the following warning:
```
odoo.fields: Redundant default on project.task.operating_unit_id
```
indeed, it’s a related field and the default value is the same on the original field.

alan196: Sorry, I didn't notice it was a related field.

alan196 (approved):
👍 Technical review

alan196:
@EduardoJarsa @hectorHerreraTeigeiro could you help me to review this PR?

EduardoJarsa (commented):
functional review everything looks working

EduardoJarsa (approved):
Functional review, everything looks working

project_operating_unit/models/project_task.py
hectorHerreraTeigeiro: Looks complete for me

PR #581, opened 2023-03-30
[15.0][FIX] account_operating_unit: split function for support expense operating unit

Saran440:
Bug:
Expense can't register payment when we config self-balance. it will error 
"Configuration error. The operating unit is mandatory for each line as the operating unit has been defined as self-balanced at company level."

it because this module get `reconciled_moves` from `payment.reconciled_bill_ids + payment.reconciled_invoice_ids` (payment from expense don't have bills)

This PR is split function for expense_operating_unit hook and change it to move on expense


AaronHForgeFlow (approved):
LGTM

ps-tubtim (approved):
LGTM

PR #593, opened 2023-05-30
[MIG] purchase_operating_unit: Migration to 16.0

tiagor-exo:
Sorry, the tests are failing because the module depends on account_operating_unit. I am in the process of migrating that module.


alan196:
This module is already being migrated in https://github.com/OCA/operating-unit/pull/595

PR #597, opened 2023-06-14
[15.0][FIX] purchase_request_operating_unit: operating unit when purchase request is created based on procurements

AaronHForgeFlow:
Enable the route MTO or dropship in a product. Then try to create a sales order for that product in another operating unit than the main operating unit. This constraint will raise: https://github.com/OCA/operating-unit/blob/15.0/purchase_request_operating_unit/model/purchase_request.py#L56, because it will try to create the purcahse request for the main operating unit, instead the operating unit of the procurement, which is the operating unit of the sales order. 

cc @ForgeFlow @BT-pcavero

### DIFF
--- a/purchase_request_operating_unit/model/__init__.py
+++ b/purchase_request_operating_unit/model/__init__.py
@@ -1,3 +1,4 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
 from . import purchase_request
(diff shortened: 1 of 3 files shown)


JordiMForgeFlow (approved):
LGTM 👍🏼

LoisRForgeFlow (approved):
LGTM!

PR #601, opened 2023-06-27
[15.0][FIX] sale_operating_unit: permission update OU in invoices

Saran440:
Step to error
1. Create user permission 'User: Own Documents Only'
2. login user sale and create sale order
3. when sale create invoice and select not `Regular invoice` (after create product already), it will error permission
Selection_012

Selection_011

Sale don't allow to write field in invoice. I add sudo() for write OU in invoice


AaronHForgeFlow (approved):

AaronHForgeFlow:
fasttrack

/ocabot merge patch

Issue #602, opened 2023-07-03
[12.0,14.0] product_operating_unit: change default behaviour of  'product.template'._get_default_category_id()


rmorant:
Conflict with odoo-pim/product_attribute_set
Detected when try to install product demo data and product_attribute_set addon is installed. The demo data is not set. 
An error like this are raised when other addon, like product_attribute_set, use this function.
```log
File "/opt/odoo/addons/product_attribute_set/models/product.py", line 31, in _get_default_att_set
  [("id", "=", default_categ_id_id.id)]
  AttributeError: 'int' object has no attribute 'id'
```

addon: product_operating_unit.

##  Bug
In all versions
'product.template' default behaviour:
```py
>>> env['product.template']._get_default_category_id()
product.category(1,)
```
When product_operating_unit addon is installed:
```py
>>> env['product.template']._get_default_category_id()
2023-07-03 07:14:47,368 1 INFO o14demo odoo.addons.product_operating_unit.models.product_template: Main Operating Unit 
1
```
Was introduced in issue #248 

**Expected behavior**
The default recordset object should be returned from _get_default_category_id()

PR #603, opened 2023-07-03
[FIX] product_operating_unit: get_default_category(): object

rmorant:
Fix Issue #602 
The default recordset object is returned from _get_default_category_id()

### DIFF
--- a/product_operating_unit/models/product_template.py
+++ b/product_operating_unit/models/product_template.py
@@ -64,7 +64,7 @@ class ProductTemplate(models.Model):
             _logger.info("%s" % (ou_id.name))
             category = self.env["product.category"].search([], limit=1)
             if category:
-                return category.id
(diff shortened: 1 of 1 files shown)


rmorant:
Hi @Saran440 , can you check this. Its a trivial change.
GiuTav can you check it in the migration.
Thank you.

Saran440 (approved):
Code Reviewed. :+1: 

Function `_get_default_category_id()` in core odoo return with object.
So, This function in `product_operating_unit` should return with object too.

However, I don't use this module in my production. Someone check?

AaronHForgeFlow:
It seems a trivial change but I don't use this module, if actual users approve the change then I will merge

AaronHForgeFlow (approved):
Merging, as clear bugfixing

PR #609, opened 2023-07-19
[13.0][IMP] account_operating_unit: added migration scripts

AaronHForgeFlow:
Operating units should be passed from invoices when migrating. This assumes the database was migrated using OpenUpgrade

@ForgeFlow @MiquelRForgeFlow

### DIFF
--- a/account_operating_unit/migrations/13.0.1.0.0/post-migration.py
+++ b/account_operating_unit/migrations/13.0.1.0.0/post-migration.py
@@ -1,0 +1,28 @@
+from openupgradelib import openupgrade
+
+
+def fill_account_move_operating_unit(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move_line aml
+        SET operating_unit_id = ail.operating_unit_id
+        FROM account_invoice_line ail
+        WHERE aml.old_invoice_line_id = ail.id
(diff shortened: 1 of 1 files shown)


account_operating_unit/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+        env.cr,
+        """
+        UPDATE account_move_line aml
+        SET operating_unit_id = ail.operating_unit_id
+        FROM account_invoice_line ail
+        JOIN account_invoice ai ON ail.invoice_id = ai.id
+        JOIN account_move am ON am.old_invoice_id = ai.id
+        WHERE am.old_invoice_id = ai.id
+        AND aml.move_id = am.id
+        AND aml.operating_unit_id is NULL""",
```
MiquelRForgeFlow:
```suggestion
        FROM account_invoice_line ail
        WHERE aml.old_invoice_line_id = ail.id
        AND aml.operating_unit_id is NULL""",
```

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow (approved):

JordiMForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

PR #582, opened 2023-03-30
[15.0][IMP] account_operating_unit: add OU in move line outstanding payment

Saran440:
Step to test:
1. Create vendor bills > confirm > register payment (register payment with journal is not set operating unit)
2. JE will not add OU in Outstanding Payment
Selection_003

This PR is add OU following:
- self-balanced and set operating unit in journal = Outstanding Payment will add OU journal
- self-balanced and not set operating unit in journal = Error
- Not self-balanced and set operating unit in journal = Outstanding Payment will add OU journal
- Not self-balanced and Not operating unit in journal = Outstanding Payment will add OU from invoices

PR #610, opened 2023-07-19
[15.0][FIX] account_operating_unit: assign operating unit of the invoice journal has no operating unit no matter if OU are self balanced

AaronHForgeFlow:
In this fix https://github.com/OCA/operating-unit/pull/582 the following criteria was implemented in regards operating units for payments:

- self-balanced and set operating unit in journal = Outstanding Payment will add OU journal
- self-balanced and not set operating unit in journal = Error
- Not self-balanced and set operating unit in journal = Outstanding Payment will add OU journal
- Not self-balanced and Not operating unit in journal = Outstanding Payment will add OU from invoices

After some testing, I thing the second criteria should not trigger an error. We should add the OU from invoices in the case OU are balanced and not operating unit in the journal. In previous versions it was possible to have a OU that is self-balanced and journals without operating unit, and the journal items for the payment had the operating unit of the invoice in that case.

@ForgeFlow

### DIFF
--- a/account_operating_unit/models/account_payment.py
+++ b/account_operating_unit/models/account_payment.py
@@ -31,13 +31,10 @@ class AccountPayment(models.Model):
         )
         invoices_ou = invoices.operating_unit_id
         if invoices and len(invoices_ou) == 1 and invoices_ou != self.operating_unit_id:
-            self_balanced = self.env.company.ou_is_self_balanced
             destination_account_id = self.destination_account_id.id
             for line in lines:
-                if (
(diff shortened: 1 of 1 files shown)


Saran440 (approved):
Functional Test, it work :+1: 
However, can you check runboat error please

MiquelRForgeFlow (approved):

AaronHForgeFlow:
Thank you @Saran440 , the runbot error was a temporary error with the runbot server. It is green now.

PR #600, opened 2023-06-22
[16.0] [MIG] sale_operating_unit

Borruso:
Migration hr_operating_unit from 15.0 to 16.0
Depends https://github.com/OCA/operating-unit/pull/596 and https://github.com/OCA/operating-unit/pull/592

--
Confermo di aver firmato il CLA https://odoo-community.org/page/cla e di aver letto le linee guida su https://odoo-community.org/page/contributing

PR #598, opened 2023-06-22
[16.0] [MIG] stock_operating_unit

Borruso:
Migration stock_operating_unit from 15.0 to 16.0

--
Confermo di aver firmato il CLA https://odoo-community.org/page/cla e di aver letto le linee guida su https://odoo-community.org/page/contributing

PR #614, opened 2023-07-27
[16.0] migration of sale_stock_operating_unit

alan196:
This PR depends on https://github.com/OCA/operating-unit/pull/600 and https://github.com/OCA/operating-unit/pull/598


francesco-ooops:
@AaronHForgeFlow @alan196 can you rebase? Runboat is not available

alan196:
@francesco-ooops @AaronHForgeFlow I rebased the branch

francesco-ooops (changes requested):
Functional review: with this module installed, the OU is loaded correctly on SO creation, but as soon as partner is set it's changed to another OU

AaronHForgeFlow (approved):
functional + light code review

AaronHForgeFlow:
Hi @francesco-ooops did you expect the OU to be changed based on the partner OU? That is not covered in this module. I am not sure if I understood.

francesco-ooops:
@AaronHForgeFlow maybe I'm missing something (and I could not reproduce the behavior on branch 16 runboat), but why is OU changing when I change SO partner? [VIDEO](https://www.loom.com/share/beca58ac91fa4fc99b07f6ad4eb19f49)

alan196:
@francesco-ooops I think your issue is caused by a method in this module, when you change the sales team, that is taken from the parter, if try to change the warehouse to a warehouse that has the same operating unit as the sales team.

francesco-ooops:
Where's the sales team in the SO partner? 

I could not reproduce the behavior in v14 or v15, that's why I flagged it

(create new OU > Create new Warehouse with new OU > Create new sales team with new OU > assign Mitchell Admin to new sales team > the NEW OU is set in SO regardless of customer selected)

francesco-ooops:
@alan196 @AaronHForgeFlow this is the difference I see between this PR and v15: [VIDEO](loom.com/share/dba5e28d41224725811dfd48dd322914?focus_title=1&muted=1&from_recorder=1)

Chionne27:
Hi @alan196 @AaronHForgeFlow. I made a new PR #797 with the migration of the module because i effectively noticed the issue reported by @francesco-ooops. The issue was caused because in v17 the modules sale and sale_stock are not properly compatible with the module sale_stock_operating_unit, that adds onchanges instead of interact with existing onchanges/depends. 
For example in v17 we have this function in sale
https://github.com/odoo/odoo/blob/e1fb54cb3d5db9c906946fa11ff49c1ca86d3722/addons/sale/models/sale_order.py#L451
and i think is better to override this function as here 
https://github.com/OCA/operating-unit/blob/745a743208b583fe79602e2012d0400f8979b9c2/sale_stock_operating_unit/models/sale_order.py#L41
instead of making an onchange and it works as it should.
Check my PR and tell me what you think

PR #616, opened 2023-08-07
[15.0][FIX] account_operating_unit: domain analytic from operating unit line

Saran440:
Step to error:
1. Create Journal Entries
2. In header not select operating unit
3. Add a line with OU1
4. It will invisible analytic with OU1

Selection_005

Analytic Account has 2 OU but can't selected
Selection_006


AaronHForgeFlow (approved):
Functional test. Fast tracking.

PR #633, opened 2023-10-26
[14.0][FIX] account_financial_report_operating_unit: filter OU trial balance

Saran440:
It will wrong amount when filter OU with trial balance (initial balance)

Add filter operating unit domain initial balance fy pl on trial balance


AaronHForgeFlow (approved):
:+1:

PR #642, opened 2024-01-04
[14.0][FIX] operating_unit: default operating unit more multi-company friendly

HekkiMelody:
This PR fixes an issue with the method operating_unit_default_get where it would return an operating unit even for a company that was not active, which would cause all sorts of issues

Steps to reproduce on runboat,
taking as example purchase_operating_unit:

- User has a default ou belonging to Company 1
- Switch to Company 2 (make sure Company 1 is not active at all)
- Try to create a Purchase Order

The issue arises because the operating_unit field is pre-compiled on the PO with an Operating Unit whose company is inactive, and the onchanges cannot find or access related data.

### DIFF
--- a/operating_unit/models/res_users.py
+++ b/operating_unit/models/res_users.py
@@ -14,7 +14,19 @@ class ResUsers(models.Model):
         if not uid2:
             uid2 = self.env.user.id
         user = self.env["res.users"].browse(uid2)
-        return user.default_operating_unit_id
+        # check if the company of the default OU is active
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
The change seems legit

francesco-ooops (approved):
Functional ok!

@AaronHForgeFlow can you review code?

ilyasProgrammer (approved):
LGTM

AaronHForgeFlow (approved):
LGTM

AaronHForgeFlow:
can you please do the forward port to newer versions?

PR #647, opened 2024-01-19
[14.0][FIX] purchase_operating_unit: default picking_type without OU

HekkiMelody:
Fixes an issue when creating a PO, if the user doesn't have any operating units assigned.

### DIFF
--- a/purchase_operating_unit/models/purchase_order.py
+++ b/purchase_operating_unit/models/purchase_order.py
@@ -14,14 +14,15 @@ class PurchaseOrder(models.Model):
         res = super(PurchaseOrder, self)._default_picking_type()
         type_obj = self.env["stock.picking.type"]
         operating_unit = self.env["res.users"].operating_unit_default_get(self.env.uid)
-        types = type_obj.search(
-            [
-                ("code", "=", "incoming"),
-                ("warehouse_id.operating_unit_id", "=", operating_unit.id),
-            ]
-        )
-        if types:
-            res = types[:1].id
+        if operating_unit:
+            types = type_obj.search(
+                [
+                    ("code", "=", "incoming"),
+                    ("warehouse_id.operating_unit_id", "=", operating_unit.id),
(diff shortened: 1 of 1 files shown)


purchase_operating_unit/models/purchase_order.py
```diff
@@ -14,14 +14,15 @@ def _default_picking_type(self):
...
-            res = types[:1].id
+        if operating_unit:
+            types = type_obj.search(
+                [
+                    ("code", "=", "incoming"),
+                    ("warehouse_id.operating_unit_id", "=", operating_unit.id),
+                ]
+            )
+            if types:
+                res = types[0].id
```
AaronHForgeFlow: Before it was `res = types[:1].id` why this change?

HekkiMelody: Seemed cleaner to me, but I can revert it.

`[:1]` is useful to make sure there are no errors on an empty recordset. But we already have `if types`, and also it would error out when trying to access `.id` anyway.

Either way, that's not the point of the issue, so I'll remove it.

francesco-ooops (approved):
Functional ok!

@AaronHForgeFlow good for you?

AaronHForgeFlow (approved):
LGTM. Thanks for the explanation

AaronHForgeFlow:
/ocabot merge patch

If you do the FW to 15.0 it would merge it right away

HekkiMelody:
Thank you @AaronHForgeFlow !

HekkiMelody:
The method I'm fixing isn't even present in v15, I'm afraid it'll require to FW multiple commits to align the versions...

francesco-ooops:
@AaronHForgeFlow can you make an analysis of commits in v14 missing in v15?

AaronHForgeFlow:
hmmm, theoretically every PR merged is forward ported to new versions. I will check when I have some time

PR #650, opened 2024-02-26
[15.0][FIX] hr_payroll_account_operating_unit: assign the operating unit to the payslip journal items

AaronHForgeFlow:
Assign the operating unit to the payslip journal items. Doing it in the write method is not sufficient when the journal item is created before. And the constraint here is raised: 

https://github.com/OCA/operating-unit/blob/15.0/account_operating_unit/models/account_move.py#L205

This depends on https://github.com/OCA/payroll/pull/146

cc @ForgeFlow


JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
LG

AaronHForgeFlow:
Merging, as without this there is an error when creating payslips journal items.

/ocabot merge patch

PR #651, opened 2024-02-28
[13.0][IMP][mis_builder_operating_unit] Filter account entries not having operating unit

robyf70:
Add explict filter for account entries without having set the operating unit"

### DIFF
--- a/mis_builder_operating_unit/model/mis_builder.py
+++ b/mis_builder_operating_unit/model/mis_builder.py
@@ -2,9 +2,10 @@
 # Jordi Ballester Alomar
 # Copyright 2015-19 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
 # Copyright 2018-19 ACSONE SA/NV
+# Copyright 2024 Level Prime Srl - Roberto Fichera
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class MisReportInstance(models.Model):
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
IMO all journal items should have an operating unit.

robyf70:
> IMO all journal items should have an operating unit.

As I know you can still post account entries without OU, so it isn't mandatory field

AaronHForgeFlow:
Yes, that is a good point. I will wait for others feedback.

AaronHForgeFlow (approved):
Code LG

AaronHForgeFlow (changes requested):
I think the filter show be added in the main instance as well. If you are not in comparison mode users should be able to filter journals by no operating unit. Can you check this? https://github.com/LevelPrime/operating-unit/pull/1

robyf70:
> I think the filter show be added in the main instance as well. If you are not in comparison mode users should be able to filter journals by no operating unit. Can you check this? [LevelPrime#1](https://github.com/LevelPrime/operating-unit/pull/1)

You are right! I will add this feature on the main instance as well

robyf70:
@AaronHForgeFlow I've just pushed the requested changes

AaronHForgeFlow (approved):
LGTM thanks

robyf70:
@AaronHForgeFlow Do we need another review to be able to merge?

mis_builder_operating_unit/model/mis_builder.py
```diff
@@ -32,16 +47,22 @@ def _get_additional_move_line_filter(self):
...
-                (
-                    "operating_unit_id",
-                    "in",
-                    sudoself.report_instance_id.operating_unit_ids.ids,
+
+        if sudoself.has_no_operating_unit:
+            aml_domain.append(("operating_unit_id", "=", False,))
+        else:
+            if sudoself.report_instance_id.has_no_operating_unit:
+                aml_domain.append(("operating_unit_id", "=", False,))
```
MiquelRForgeFlow:
```suggestion
        elif sudoself.report_instance_id.has_no_operating_unit:
            aml_domain.append(("operating_unit_id", "=", False,))
```

mis_builder_operating_unit/model/mis_builder.py
```diff
@@ -32,16 +47,22 @@ def _get_additional_move_line_filter(self):
...
+            elif sudoself.report_instance_id.operating_unit_ids:
```
MiquelRForgeFlow:
```suggestion
        else:
            if sudoself.report_instance_id.operating_unit_ids:
```

mis_builder_operating_unit/model/mis_builder.py
robyf70: Ok! Done!

mis_builder_operating_unit/model/mis_builder.py
robyf70: Ok! Done!

robyf70:
@MiquelRForgeFlow Are you able to review the last changes so we can finally get it merged?

MiquelRForgeFlow (approved):

PR #658, opened 2024-04-24
[15.0][IMP] account_operating_unit: allow to pay multiple invoices with different operating units

AaronHForgeFlow:
That was possible in old versions <v12 (https://github.com/OCA/operating-unit/blob/12.0/account_operating_unit/models/account_payment.py#L71) but it was removed for unknown reason when migrating to v13.

In general, reconciling a bank journal entry with an invoice will require Inter OU balancing entries, as long one OU is paying the balance for the other.

If needed I can add a company parameter for restricting paying invoices for different OU if you want to keep compatibility with the actual code.

@Saran440 @alan196 @ForgeFlow


alan196 (approved):
LGTM

alan196:
@GasparJarsa check this PR

GasparJarsa (approved):
LGFM

account_operating_unit/models/account_move.py
```diff
@@ -65,6 +65,63 @@ def _check_move_operating_unit(self):
...
+            return super().reconcile()
+        bank_journal = bank_journal[0]
+        # If all move lines point to the same operating unit, there's no
+        # need to create a balancing move line
+        if len(self.mapped("operating_unit_id")) <= 1:
+            return super().reconcile()
+        # Create balancing entries for un-balanced OU's.
+        move_vals = self._prepare_inter_ou_balancing_move(bank_journal)
+        move = self.env["account.move"].create(move_vals)
+        ou_balances = self._check_ou_balance(self)
```
MiquelRForgeFlow, line 96:
```suggestion
        ou_balances = self.move_id._check_ou_balance(self.move_id)
```

MiquelRForgeFlow: and delete above _check_ou_balance

account_operating_unit/models/account_move.py
```diff
@@ -65,6 +65,63 @@ def _check_move_operating_unit(self):
...
+            line_data = move._prepare_inter_ou_balancing_move_line(
+                move, ou_id, ou_balances
+            )
+            if line_data:
+                amls.append(self.with_context(wip=True).create(line_data))
+        if amls:
+            move.with_context(wip=False).write(
+                {"line_ids": [(4, aml.id) for aml in amls]}
+            )
+        move.with_context(inter_ou_balance_entry=True)._post()
```
MiquelRForgeFlow:
```suggestion
        move.with_context(inter_ou_balance_entry=True).action_post()
```
maybe?

account_operating_unit/models/account_move.py
AaronHForgeFlow: This is not the same code. In the reconcile method I am calling that method with 2 separate account.move.line records from different account moves.

account_operating_unit/models/account_move.py
AaronHForgeFlow: ok, done that.

MiquelRForgeFlow (approved):

PR #662, opened 2024-05-14
Deprecate the onchange return domain

Ppamungkas:
Odoo 14 deprecates domain return on onchange function.


AaronHForgeFlow:
Thanks for the fix. Just please leave one commit only, and remove the commented unused code.

PR #664, opened 2024-05-15
[14.0][FIX] purchase_operating_unit: Fix warning from an onchange function returning domain. Use domain fi…

Ppamungkas:
…eld instead.

### DIFF
--- a/purchase_operating_unit/__manifest__.py
+++ b/purchase_operating_unit/__manifest__.py
@@ -20,6 +20,7 @@
         "report/purchase_report_view.xml",
         "views/purchase_order_view.xml",
         "views/purchase_order_line_view.xml",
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Can you please include the module name in the commit message? for example:

`[FIX] purchase_operating_unit: Fix warning from an onchange function returning domain. Use domain field instead.`

Ppamungkas:
@AaronHForgeFlow sorry for the mess, I just need to get a hang on it with trial and error...

AaronHForgeFlow (approved):
@Ppamungkas  thanks for attend the comments. I think this is good :+1:

MiquelRForgeFlow (approved):

PR #667, opened 2024-05-28
[MIG] operating_unit: Migration to 17.0

ghost:
Compare to 16.0, I added a new method in the model "res.users". The method is inspired by the one on companies (https://github.com/odoo/odoo/blob/7266a662171fedc945523c5bc1bb34a4dd0995bf/odoo/api.py#L645) in the idea that at some point maybe a selector widget (like company one) will be added to choose the active OUs


AaronHForgeFlow (commented):
Code + functional LG. Just please notice that tests fail because of this:

```
  File "/__w/operating-unit/operating-unit/operating_unit/tests/test_operating_unit.py", line 153, in test_find_operating_unit_by_name_or_code
    self.assertEqual(ou.name_search("nam"), expected_result)
AssertionError: Lists differ: [(10, 'name')] != [(10, '[code] name')]

First differing element 0:
(10, 'name')
(10, '[code] name')

- [(10, 'name')]
+ [(10, '[code] name')]
?        +++++++
```

probably something with the display_name?

map-240221:
Is there any reason why the record rules was changed from:

```
    <record id="ir_rule_operating_unit_allowed_operating_units" model="ir.rule">
        <field name="model_id" ref="model_operating_unit" />
        <field name="domain_force">[('id','in',user.operating_unit_ids.ids)]</field>
```

To this:

```
    <record id="ir_rule_operating_unit_allowed_operating_units" model="ir.rule">
        <field name="model_id" ref="model_operating_unit" />
        <field name="domain_force">[('id','in',operating_unit_ids)]</field>
```

With the later, the rule doesn't seem to properly take effect (i.e. all operating units are shown instead of only the allowed ones)

ghost:
> Code + functional LG. Just please notice that tests fail because of this:
> 
> ```
>   File "/__w/operating-unit/operating-unit/operating_unit/tests/test_operating_unit.py", line 153, in test_find_operating_unit_by_name_or_code
>     self.assertEqual(ou.name_search("nam"), expected_result)
> AssertionError: Lists differ: [(10, 'name')] != [(10, '[code] name')]
> 
> First differing element 0:
> (10, 'name')
> (10, '[code] name')
> 
> - [(10, 'name')]
> + [(10, '[code] name')]
> ?        +++++++
> ```
> 
> probably something with the display_name?

@AaronHForgeFlow Yeah I missed a name_get and replaced it. 



> Is there any reason why the record rules was changed from:
> 
> ```
>     <record id="ir_rule_operating_unit_allowed_operating_units" model="ir.rule">
>         <field name="model_id" ref="model_operating_unit" />
>         <field name="domain_force">[('id','in',user.operating_unit_ids.ids)]</field>
> ```
> 
> To this:
> 
> ```
>     <record id="ir_rule_operating_unit_allowed_operating_units" model="ir.rule">
>         <field name="model_id" ref="model_operating_unit" />
>         <field name="domain_force">[('id','in',operating_unit_ids)]</field>
> ```
> 
> With the later, the rule doesn't seem to properly take effect (i.e. all operating units are shown instead of only the allowed ones)

You should not have any functional differences. I simply added the keyword operating_unit_ids to the ir_rules and a method to the model res_users.

```
    @api.model
    def _eval_context(self):
        res = super()._eval_context()
        res.update(
            {
                "operating_unit_ids": self.env.user.operating_units().ids,
                "operating_unit_id": self.env.user.default_operating_unit_id.id,
            }
        )
        return res       
```

```
    def operating_units(self):
        return self.env.user.operating_unit_ids        
```

operating_unit/models/operating_unit.py
```diff
@@ -41,14 +41,10 @@ class OperatingUnit(models.Model):
...
-    def name_get(self):
-        res = []
+    @api.depends("name", "code")
+    def _compute_display_name(self):
         for ou in self:
-            name = ou.name
-            if ou.code:
-                name = f"[{ou.code}] {name}"
-            res.append((ou.id, name))
-        return res
```
vvrossem: This change should be in the migration commit

operating_unit/models/res_users.py
```diff
@@ -29,51 +53,32 @@ def operating_unit_default_get(self, uid2=False):
 
     @api.model
     def _default_operating_unit(self):
-        return self.operating_unit_default_get()
+        return self._get_default_operating_unit()
 
     @api.model
     def _default_operating_units(self):
         return self._default_operating_unit()
```
vvrossem: Is it possible to use a single method instead ?

vvrossem (commented):
You can squash your last commit into the migration commit

operating_unit/security/operating_unit_security.xml
```diff
@@ -27,7 +27,7 @@
     </record>
     <record id="ir_rule_operating_unit_allowed_operating_units" model="ir.rule">
         <field name="model_id" ref="model_operating_unit" />
-        <field name="domain_force">[('id','in',user.operating_unit_ids.ids)]</field>
+        <field name="domain_force">[('id','in',operating_unit_ids)]</field>
```
vvrossem, line 30: Why is `user` removed?

ghost: The keyword "operating_unit_ids" is now in the context of the ir.rules
The idea is to prepare the ground for a potential widget selector (like the company one) and to have a method governing the whole thing. 

You can see the changes related to that in ir_rule.py and res_users.py (def operating_units)

operating_unit/models/res_users.py
ghost: I think so. Code updated.
Thanks for the review 👍

operating_unit/tests/common.py
```diff
@@ -0,0 +1,68 @@
...
+        user = cls.res_users_model.create(
+            {
+                "name": "Test User",
+                "login": login,
+                "password": "demo",
+                "email": "test@yourcompany.com",
+                "company_id": company.id,
+                "company_ids": [(4, company.id)],
+                "operating_unit_ids": [(4, ou.id) for ou in operating_units],
+                "default_operating_unit_id": False,
```
yankinmax: @jdidderen-nsi can you pls remove this value, because tests in `res_partner_operating_unit` are failing.
Why actually you want to set `default_operating_unit_id` to False? Is there any reason?

operating_unit/models/operating_unit.py
```diff
@@ -42,22 +41,18 @@ class OperatingUnit(models.Model):
...
-        res = []
+    @api.depends("name", "code")
+    def _compute_display_name(self):
         for ou in self:
-            name = ou.name
-            if ou.code:
-                name = "[{}] {}".format(ou.code, name)
-            res.append((ou.id, name))
-        return res
+            ou.display_name = f"[{ou.code}] {ou.name}"
```
Camille0907, line 47: In the previous version, we were managing the case where `ou.code` is empty -> we don't want to keep it ?

pandoc-3.2-1-amd64.deb
Camille0907, line 1: Why do we need to add this file ?

Camille0907 (commented):
For another time, maybe it would be a bit better to split in two commits what's needed for the migration / what's an improvement, it would ease the review :) 
Thank you !

pandoc-3.2-1-amd64.deb
yankinmax: it's auto added with `pre-commit`

operating_unit/models/operating_unit.py
ghost: The code fields is required on the operating unit so we should not have this case

operating_unit/tests/common.py
yankinmax: @jdidderen-nsi 
`"default_operating_unit_id": False,` is wrong value for user being created, please remove that and let Odoo assign real default.

ghost: @yankinmax The piece of code is removed

yankinmax (approved):

vvrossem (approved):

pandoc-3.2-1-amd64.deb
gurneyalex: This sounds wrong. 

I opened https://github.com/OCA/maintainer-tools/issues/619 to check this.

gurneyalex: OK so it must be removed: https://github.com/OCA/oca-addons-repo-template/issues/262

gurneyalex, line 1: please remove this file

gurneyalex (changes requested):

AaronHForgeFlow (commented):
Thank you for the migration and the refactor. Functional tests are good so far. Tested it using the migration of the account_operating_unit module for it. When the uneeded file is removed then I am fine with a merge.

pandoc-3.2-1-amd64.deb
ghost: And the file is removed 👍

yankinmax:
@AaronHForgeFlow @gurneyalex 
Can you pls help to get this merged?

gurneyalex (approved):

ghost:
Just a quick message to thanks everyone for the help and the reviews on these modules.

FYI I will probably work on operating units during OCA days. For now, I have two improvements :

- Have a selector widget in the same approach as multi-company
- Improve report_qweb_operating_unit to have a different layout per operating unit (e.g having boxed in OU1 and stripped on OU2)

PR #668, opened 2024-05-28
[MIG] product_operating_unit: Migration to 17.0

ghost:
Some changes (security rules) are based on changes made in the base modules https://github.com/OCA/operating-unit/pull/667

I removed domains on operating units fields for two reasons:
- A operating unit manager is not necessarily assigned to a operating unit (https://github.com/OCA/operating-unit/blob/025e778a91e56bb83a52ec46dde19438965ed12c/operating_unit/models/res_users.py#L76) so in this case, the user id is not present in the user_ids field of the operating unit
- The security rules already filter the units I can see.

Depends on:


product_operating_unit/models/product_category.py
```diff
@@ -0,0 +1,30 @@
...
+        if vals.get("operating_unit_ids"):
+            for rec in self:
+                products = product_template_obj.search(
+                    [("categ_id", "child_of", rec.id)]
+                )
+                category_ou_ids = rec.operating_unit_ids.ids
+                for product in products:
+                    ou_ids = product.operating_unit_ids.ids
+                    ou_ids.extend(category_ou_ids)
+                    product.operating_unit_ids = [Command.set(ou_ids)]
```
SilvioC2C: Nitpicking: might as well use records directly, instead of ids

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,69 @@
...
+        default=lambda self: self._default_operating_unit_ids(),
+    )
+    categ_id = fields.Many2one(
+        default=lambda self: self._get_default_category_id(),
+    )
+
+    @api.model
+    def _default_operating_unit_ids(self):
+        if self.categ_id and self.categ_id.operating_unit_ids:
+            return [(6, 0, self.categ_id.operating_unit_ids.ids)]
```
SilvioC2C: Since this method is used as a default method, `self.categ_id` will always be empty.

product_operating_unit/models/product_template.py
```diff
@@ -0,0 +1,69 @@
...
+                    _(
+                        "The operating units of the product must include the "
+                        "ones from the category."
+                    )
+                )
+
+    @api.depends("categ_id")
+    def _compute_operating_unit_ids(self):
+        for record in self:
+            if record.categ_id.operating_unit_ids:
```
SilvioC2C: I think you can remove this `if` statement: if the category is changed to a category with no OU, the field can be safely emptied

SilvioC2C (commented):
Small improvements suggested.
CI is red, though.

product_operating_unit/models/product_category.py
yankinmax: at least it's worth it to try if it fails on this

yankinmax:
you need to add `test-requirements.txt` file in `operating-unit` folder with such content
```
odoo-addon-operating_unit @ git+https://github.com/OCA/operating-unit.git@refs/pull/667/head#subdirectory=operating_unit
```
this should fix the build

product_operating_unit/models/product_category.py
ghost: Logic updated with records instead of ids. Thanks for the review 👍

product_operating_unit/models/product_template.py
ghost: For now, I simply removed the if statement but I'm wondering if the default is still needed with the compute based on the categ_id

product_operating_unit/models/product_template.py
ghost: Statement removed 👍

yankinmax (approved):

yankinmax:
@jdidderen-nsi thanks for the migration

product_operating_unit/models/product_template.py
```diff
@@ -3,23 +3,29 @@
...
     _inherit = "product.template"
 
+    operating_unit_ids = fields.Many2many(
+        "operating.unit",
+        "product_operating_unit_rel",
+        string="Operating Units",
+        compute="_compute_operating_unit_ids",
+        store=True,
+        readonly=False,
+        default=lambda self: self._default_operating_unit_ids(),
```
Camille0907: Do you still need the default, with the compute ?

Camille0907 (approved):
Migration LGTM, thanks

product_operating_unit/models/product_template.py
ghost: This was one of my thoughts, given the compute based on the categ_id
So piece of code is removed

gurneyalex:
#667 is now merged :smile_cat:

yankinmax:
@jdidderen-nsi can you please remove dependency commit and rebase?
operating_unit is merged:

ghost:
@gurneyalex @yankinmax Rebased and good to go 👍

PR #671, opened 2024-05-29
[MIG] sales_team_operating_unit: Migration to 17.0

ghost:
Depends on:

- [x] https://github.com/OCA/operating-unit/pull/667

PR #670, opened 2024-05-29
[MIG] account_operating_unit: Migration to 17.0

ghost:
Depends on:

- [x] https://github.com/OCA/operating-unit/pull/667
- [x] https://github.com/OCA/operating-unit/pull/669

PR #669, opened 2024-05-28
[MIG] analytic_operating_unit: Migration to 17.0

ghost:
Depends on:

- [x] https://github.com/OCA/operating-unit/pull/667

PR #672, opened 2024-05-29
[MIG] sale_operating_unit: Migration to 17.0

ghost:
Depends on:

- [x] https://github.com/OCA/operating-unit/pull/669
- [x] https://github.com/OCA/operating-unit/pull/670
- [x] https://github.com/OCA/operating-unit/pull/671


yankinmax (approved):

sale_operating_unit/tests/test_sale_operating_unit.py
```diff
@@ -2,56 +2,58 @@
...
-        cls.user2 = cls._create_user(
-            "user_2", [cls.grp_sale_user, cls.grp_acc_user], cls.company, [cls.b2c]
+        cls.user2.write(
+            {
+                "groups_id": [
+                    Command.link(cls.grp_sale_user.id),
+                    Command.link(cls.grp_acc_user.id),
+                ],
+                "operating_unit_ids": [
+                    Command.set(cls.b2c.id),
```
yankinmax: `set` isn't the same as `link`
so I think you might be able to link like this `[Command.link(ou.id) for ou in operating_units],` if there are multiple units or just `Command.link(cls.b2c.id)`

sale_operating_unit/models/sale_order.py
```diff
@@ -16,6 +16,7 @@ class SaleOrder(models.Model):
         readonly=False,
         store=True,
         precompute=True,
+        check_company=True,
```
yankinmax, line 18: nice improvement, thanks

sale_operating_unit/tests/test_sale_operating_unit.py
ghost: Since the b2c OU is already declared for user2 in the common class, this piece of code is not useful anymore.

sale_operating_unit/models/sale_order.py
```diff
@@ -9,30 +9,59 @@
...
 
+    operating_unit_id = fields.Many2one(
+        comodel_name="operating.unit",
+        string="Operating Unit",
+        default=lambda self: self._default_operating_unit(),
+        readonly=False,
+        store=True,
+        precompute=True,
+        check_company=True,
+        compute="_compute_operating_unit_id",
```
Camille0907, line 19: Why do you need a `default` method, since you have a `compute` one ?

Camille0907 (approved):
Technically LGTM :)

sale_operating_unit/models/sale_order.py
```diff
@@ -9,30 +9,59 @@
...
 
-    @api.onchange("operating_unit_id")
-    def onchange_operating_unit_id(self):
-        if self.team_id and self.team_id.operating_unit_id != self.operating_unit_id:
-            self.team_id = False
+    @api.depends("operating_unit_id")
+    def _compute_journal_id(self):
+        res = super()._compute_journal_id()
+        for sale in self:
+            sale.journal_id = (
```
nguyenminhchien: This overwrites the one we have in `sale_management`: https://github.com/odoo/odoo/blob/17.0/addons/sale_management/models/sale_order.py#L73
```
    @api.depends('sale_order_template_id')
    def _compute_journal_id(self):
        super()._compute_journal_id()
        for order in self.filtered('sale_order_template_id'):
            order.journal_id = order.sale_order_template_id.journal_id

```

sale_operating_unit/models/sale_order.py
```diff
@@ -77,3 +90,12 @@ class SaleOrderLine(models.Model):
         string="Operating Unit",
         store=True,
     )
+
+    def _prepare_invoice_line(self, **optional_values):
+        values = super()._prepare_invoice_line(**optional_values)
+        values.update(
+            {
+                "operating_unit_id": self.operating_unit_id.id,
```
nguyenminhchien: you don't need to set "operating_unit_id" for account.move.line. it will be handled by: https://github.com/OCA/operating-unit/blob/95d72a43f22c1c6c87228cd4b4339268026a16ff/account_operating_unit/models/account_move.py#L18

image

sale_operating_unit/views/sale_view.xml
```diff
@@ -39,7 +39,23 @@
...
+            </field>
+            <field name="analytic_account_id" position="attributes">
+                <attribute
+                    name="domain"
+                >["|",("operating_unit_ids","in", operating_unit_id),("operating_unit_ids","in", False)]</attribute>
+            </field>
+            <field name="journal_id" position="attributes">
+                <attribute
+                    name="domain"
+                >["|",("operating_unit_id","=", operating_unit_id),("operating_unit_id","=", False)]</attribute>
```
nguyenminhchien: You need to keep the domain define in module `sale` as well: `('type', '=', 'sale')`

sale_operating_unit/views/sale_view.xml
```diff
@@ -39,7 +39,23 @@
...
+                <field
+                    name="operating_unit_id"
+                    invisible="1"
+                    groups="!operating_unit.group_multi_operating_unit"
+                />
+            </field>
+            <field name="analytic_account_id" position="attributes">
+                <attribute
+                    name="domain"
+                >["|",("operating_unit_ids","in", operating_unit_id),("operating_unit_ids","in", False)]</attribute>
```
nguyenminhchien: I face this warning:
`2024-06-18 09:18:06,328 141803 WARNING v17c_sale_operating_unit odoo.osv.expression: The domain term '('operating_unit_ids', 'in', False)' should use the '=' or '!=' operator. `

nguyenminhchien: on the other hand, you should keep the domain defined in python file
`['|', ('company_id', '=', False), ('company_id', '=', company_id)]`

nguyenminhchien (changes requested):

sale_operating_unit/models/sale_order.py
ghost: Code removed 👍

sale_operating_unit/models/sale_order.py
ghost: I updated the logic around the compute to only update when the journal is not in the defined operating unit.

sale_operating_unit/models/sale_order.py
ghost: Code removed 👍

sale_operating_unit/views/sale_view.xml
ghost: Fixed 👍

sale_operating_unit/views/sale_view.xml
ghost: I fixed the domain itself. 
I also added the domain defined in the python file but I'm not sure it's useful. I mean the domain on the company_id is automatically added by the check_company

sale_operating_unit/models/sale_order.py
```diff
@@ -9,30 +9,59 @@
...
+                            ("operating_unit_id", "=", sale.operating_unit_id.id),
+                            ("operating_unit_id", "=", False),
+                            "|",
+                            ("company_id", "=", sale.company_id.id),
+                            ("company_id", "=", False),
+                            ("type", "=", "sale"),
+                        ],
+                        limit=1,
+                    )
+                    .id
```
nguyenminhchien, line 62: Non-blocking, but no need to get its `id`.

sale_operating_unit/views/sale_view.xml
nguyenminhchien: yes, it's useful. you can define the domain in python file or in the view. in my understanding, the one from the view is higher priority than the one from python file. `check_company` attribute will be handled as a domain from python file, thus, will be ignored in this case.

nguyenminhchien (approved):

yankinmax:
@AaronHForgeFlow can you pls trigger a merge here, PR seems to be finished and approved.

PR #673, opened 2024-05-29
[MIG] stock_operating_unit: Migration to 17.0

ghost:
Depends on:


ghost:
**Open discussion for the constraint on locations : _check_required_operating_unit**

In previous versions of the module, the operating unit was mandatory on locations (except customer and supplier ones). In the case of one of my projects, the customer wants to have a "central" warehouse common to all operating units. With the constraint, I have to define an operating unit, even if it's not coherent with the business need. 

Right now, I remove this part of the code from the constraint, but maybe there is another solution. 

What's the opinion on this?

AaronHForgeFlow:
I am not completely sure, but I think the condition is there mainly for correctly inventory accounting, so to ensure that the inventory value is owned by a specific operating unit. If that is the case maybe we can move the contraint to the stock_account_operating_unit module.

Another solution is to consider using hierarchy in operating units: https://github.com/OCA/operating-unit/pull/319 like if the parent OU represents the company. This approach is more complex because all the contraints definitions should be refactored.

stock_operating_unit/view/stock.xml
```diff
@@ -100,8 +98,8 @@
                 <field
                     name="operating_unit_id"
                     options="{'no_create': True}"
-                    domain="[('company_id','=', company_id)]"
                     groups="operating_unit.group_multi_operating_unit"
+                    readonly="state == 'draft'"
```
nguyenminhchien: readonly when "state **_!=_** 'draft'"

nguyenminhchien (changes requested):
Please take this occasion to  squash administrative commits (if any) with the previous commit for reducing commit noise.
Squash your 2 commits with [[MIG] stock_operating_unit: Migration to 17.0](https://github.com/OCA/operating-unit/pull/673/commits/d50283c9e1d9ec4a20195fb966aa8e5603c40958)
image

Thanks,

stock_operating_unit/view/stock.xml
ghost: Fixed 👍

nguyenminhchien (approved):

yvaucher (approved):
LGTM

Issue #617, opened 2023-08-21
account_operating_unit: inter ou balance lines are displayed in the invoice


gurneyalex:
## Module
account_operating_unit

## Describe the bug

When I use the inter operating balance feature, the module adds lines to account move which are displayed as invoice lines. They get printed out in the PDF report. 

## To Reproduce
**Affected versions**: 15.0 at least

1. set up inter operating unit balance on the company, and a balance account
2. configure the sales journal to have no operating unit
3. create a customer invoice with no operating unit, and 2 products on two different lines
4. go to the "journal items" tab, set the first product line on OU B2B, for instance, the second product line on OU B2C and the account receivable and tax lines ou the main OU
5. save and post the invoicee

**Expected behavior**
the balance lines are not displayed as an invoice lines, they are only visible in the journal items tab. They are also not printed in the PDF report. 

Current status:

![image](https://github.com/OCA/operating-unit/assets/2032807/7ec24c89-e360-4cc5-aba8-d2aeb2fcae6e)

**Additional context**

It would be also super cool to have this working with an OU on the account move, and only generate the balancing lines for the lines which don't match that OU.

AaronHForgeFlow:
That is too bad. I think that marking them as "Exclude from invoice tab" would be enough.

gurneyalex:
I came accross this testing the 16.0 migration PR, and was able to reproduce on 15.0, however I'm not sure if the way the field that filters the invoice lines is computed is the same on both versions. I need to check...

AaronHForgeFlow:
v16 is also affected

AaronHForgeFlow:
I think we can close this. It is fixed in v15 and there are PRs to fix it in v16 and v17:

https://github.com/OCA/operating-unit/pull/708

https://github.com/OCA/operating-unit/pull/739

PR #705, opened 2024-09-02
[15.0][FIX] account_operating_unit: do not show inter-OU balance journals as invoice lines

AaronHForgeFlow:
Please do merge without bumping the version number

Fixes: https://github.com/OCA/operating-unit/issues/617

I am not sure about adding a new boolean for this purpose:

- Normally companies won't allow setting posting entries back to draft
- The use case is not common, so adding a new column in a big table may not be good

We can delete the journal items by filtering the ones with name "OU-Balancing" or ref "Inter OU Balancing", but that criteria is not exact as long that reference could be added manually by the user, and the system will be deleting those journal items incorrectly.

@ForgeFlow

### DIFF
--- a/account_operating_unit/migrations/15.0.1.3.0/pre-migration.py
+++ b/account_operating_unit/migrations/15.0.1.3.0/pre-migration.py
@@ -1,0 +1,31 @@
+# Copyright 2024 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(env.cr, "account_move_line", "is_ou_balance"):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            ALTER TABLE account_move_line
+            ADD COLUMN IF NOT EXISTS is_ou_balance BOOLEAN DEFAULT false
+            """,
+        )
+        openupgrade.logged_query(
+            env.cr,
+            """ALTER TABLE account_move_line ALTER COLUMN is_ou_balance DROP DEFAULT""",
+        )
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_move_line aml0 SET is_ou_balance = true
+            FROM account_move_line aml
+            INNER JOIN res_company rc ON rc.id = aml.company_id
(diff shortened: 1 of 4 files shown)


account_operating_unit/models/account_move.py
```diff
@@ -12,6 +12,7 @@ class AccountMoveLine(models.Model):
     operating_unit_id = fields.Many2one(
         comodel_name="operating.unit",
     )
+    is_ou_balance = fields.Boolean()
```
JordiMForgeFlow:
```suggestion
    is_ou_balance = fields.Boolean(readonly=True)
```

Just to be safe :) The rest LGTM, I do not see other way to identify them without the flag.

Would be nice to have a migration script to fill old values

AaronHForgeFlow: Nice suggestions :) added both of them

JordiMForgeFlow (approved):
code review 👍🏼

account_operating_unit/migrations/15.0.1.3.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+def migrate(env, version):
+    if not openupgrade.column_exists(env.cr, "account_move_line", "is_ou_balance"):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            ALTER TABLE account_move_line
+            ADD COLUMN IF NOT EXISTS is_ou_balance BOOLEAN DEFAULT false
+            """,
+        )
+        openupgrade.logged_query(
```
MiquelRForgeFlow, line 16:
```suggestion
        openupgrade.logged_query(
                env.cr,
                """ALTER TABLE account_move_line ALTER COLUMN is_ou_balance DROP DEFAULT""",
        )
        openupgrade.logged_query(
```

AaronHForgeFlow: Done. Thanks

MiquelRForgeFlow (approved):

AaronHForgeFlow:
If no more comments are added here I will be merging this soon, better to fix the bug as soon as possible.

AaronHForgeFlow:
I think 3 weeks is enough time. We can merge this now

/ocabot merge nobump

AaronHForgeFlow:
I will propose the forward port very soon. It is not easy, because there was a refactor in journal items boolean fields between v15 & v16

PR #711, opened 2024-10-01
[17.0][FIX] access_all modules - uniformize write function on security rules

jdidderen:


AaronHForgeFlow:
LGTM if separate commit per module

jdidderen:
> LGTM if separate commit per module

@AaronHForgeFlow Done 👍

PR #713, opened 2024-10-05
[17.0][IMP] res_partner_operating_unit - remove required on operating_unit_ids

jdidderen:
In my opinion, the field operating_unit_ids should not be required on the partner especially since the security rules are fine to not have operating units


AaronHForgeFlow:
I think it makes sense

jdidderen:
@yankinmax Since you were the one who migrated the module in 17, do you have an opinion on this ?

yankinmax (approved):
I agree

PR #718, opened 2024-10-15
[17.0][FIX] crm_operating_unit: warning in logs

gurneyalex:
When crm_operating_unit is installed, warning messages are sent to the Odoo logs:

WARNING odoodb_test odoo.fields: Redundant default on crm.lead.operating_unit_id

This is caused by the setting of a default value on a related field.

We change the field to be computed, read-only with a compute method which implements the logic of the default value in case the team of the lead does not have an operating unit.

### DIFF
--- a/crm_operating_unit/models/crm_lead.py
+++ b/crm_operating_unit/models/crm_lead.py
@@ -7,16 +7,22 @@ from odoo import api, fields, models
 class CRMLead(models.Model):
     _inherit = "crm.lead"
 
-    @api.model
-    def _get_default_operating_unit(self):
-        team = self.env["crm.team"]._get_default_team_id()
-        if team.operating_unit_id:
-            return team.operating_unit_id
-        return self.env["res.users"].operating_unit_default_get(self._uid)
-
     operating_unit_id = fields.Many2one(
         "operating.unit",
         "Operating Unit",
-        related="team_id.operating_unit_id",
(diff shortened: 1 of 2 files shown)


yankinmax (approved):

AaronHForgeFlow (approved):
LGTM

crm_operating_unit/models/crm_lead.py
```diff
@@ -7,16 +7,22 @@
...
     operating_unit_id = fields.Many2one(
         "operating.unit",
         "Operating Unit",
-        related="team_id.operating_unit_id",
-        default=_get_default_operating_unit,
+        compute="_compute_operating_unit",
+        store=True,
     )
+
+    @api.depends("team_id.operating_unit_id")
```
AaronHForgeFlow, line 17: shouldn't we add the team_id also in the api.depends?

rvalyi:
@gurneyalex , please just add the team_id also in the depends as @AaronHForgeFlow suggested as changing the team might also change the operating unit...

rvalyi (changes requested):
@gurneyalex , please just add the team_id also in the depends as @AaronHForgeFlow suggested as changing the team might also change the operating unit...

vvrossem:
superseded by https://github.com/OCA/operating-unit/pull/826

PR #724, opened 2025-01-02
[IMP] Added stock_account_operating_unit to keep consisting operating units in inv. adjustments

BT-dherreros:
When a inventory adjustment is created for a real time valued product the resulting journal entry doesn't have a operating unit set (even when the underlying stock.quant has a OU set).
I created this module to fix that issue, but also as a base for any other improvements that can be done over stock_account regarding operating units.


AaronHForgeFlow:
There is a module with that name in v15. Can you please migrate that one instead?

BT-rmartin:
Please, close it. Replaced by #725

BT-dherreros:
This functionality should be added to stock_account_operating_unit after it is merged into 17.0: https://github.com/OCA/operating-unit/pull/725

PR #725, opened 2025-01-03
[MIG] stock_account_operating_unit: Migration to 17.0

BT-dherreros:


BT-rmartin (approved):
LGTM

stock_account_operating_unit/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
BT-rmartin: Wrong header, should be LGPL

AaronHForgeFlow (approved):
Functional review LGTM

BT-dherreros:
> Functional review LGTM

I had another functional look at this code and it seems to be creating 2 journal entries, one that seems to be correct and comes form Odoo's core and the other one coming from this migrated module with an incorrect ref field. Could you confirm this? maybe this migration is not needed anymore

AaronHForgeFlow (changes requested):
Yes. I think it is a past improvement done in a previous version:

```
Create stock moves between internal locations within the same company, but
different OU’s. The journal entries are created and they are self-balanced
within the OU when the journal entries are posted
```

This may be creating double JE when the products are actually moved out of the company. In my opinion, it is best to remove the action_done method, or at least, check if the products are move between internal locations of different OUs for doing the change.

I've some issues with moving products between different OU. That is why I have forbidden this action in my instances and I force to move first out of the company from OU1 and then move into the company using OU2. So, I am ok with removing the overwrite in the action_done method of stock.move

BT-dherreros:
@AaronHForgeFlow I have removed the action_done inherited method. I also adapted the unit tests that were failing. Do you think my changes to the tests make sense? To me it does of course, as the balances must change with different amount of moves.

stock_account_operating_unit/tests/test_stock_account_operating_unit.py
```diff
@@ -346,15 +340,15 @@ def test_pickings(self):
...
-        # GL account ‘Inventory’ has balance 0 on OU main_operating_unit
-        expected_balance = 0.0
+        # GL account ‘Inventory’ has balance 1 on OU main_operating_unit
+        expected_balance = 1.0
         self._check_account_balance(
             self.account_inventory.id,
             operating_unit=self.ou1,
             expected_balance=expected_balance,
         )
         # GL account ‘Inventory’ has balance 2 on OU b2c
```
AaronHForgeFlow, line 350: In my opinion the comment is correct, it should be 1 (the previous reception) + 1 (this internal transfer).

So the removal of the action_done method probably made this to have balance 1 in b2c instead of 2. Perhaps we should not remove the method but check we are not moving from in/out the company to avoid the double posting. So the override of the action_done method would be only for this inter-OU transfers between internal locations. What do you think?

stock_account_operating_unit/tests/test_stock_account_operating_unit.py
```diff
@@ -0,0 +1,363 @@
...
+            self.account_grni.id, operating_unit=None, expected_balance=expected_balance
+        )
+        # GL account ‘Goods Received Not Invoiced’ has balance -1 on Main OU
+        expected_balance = -1.0
+        self._check_account_balance(
+            self.account_grni.id,
+            operating_unit=self.ou1,
+            expected_balance=expected_balance,
+        )
+        # GL account ‘Goods Received Not Invoiced’ has balance 0 on OU b2c
```
AaronHForgeFlow, line 318: balance -1

AaronHForgeFlow (commented):
Thank you for attending the comments @BT-dherreros I still have some doubts in regards inter OU transfers. Can you please give your opinion?

stock_account_operating_unit/tests/test_stock_account_operating_unit.py
BT-dherreros: I'd say you are most probably right. I guess the best approach is readding the action_done and then checking that we are only applying it to the inter-ou internal transfers as you mentioned

PR #733, opened 2025-01-13
[ADD] account_operating_unit: add operating unit in account analytic line

alan196:
This PR adds the field operating unit to account analyitc line model and allow to generate analytic report by operating unit


AaronHForgeFlow:
It seems to make sense, I am also wondering if there should be a constraint in the analytic_operating_unit module that prevents analytic lines with different operating unit than the ones in the analytic account. If this move forward it would be awesome to have a migration script.

PR #736, opened 2025-01-24
[FIX] *_operating_unit: add check_company in field OU

Saran440:
Step to test:
1. Add multi company and create OU
Selection_019

2. Go to invoice, po, so, exp or else > select Operating Unit field it will show all OU
Selection_020

This document created on `YourCompany` it should select OU that depend on `YourCompany` only
This PR fixed by add `check_company=True` in header field.

Exclude: line, compute field, wizard (may be it should add too. but I didn't test)

After add it
Selection_021

### DIFF
--- a/account_asset_operating_unit/models/account_asset.py
+++ b/account_asset_operating_unit/models/account_asset.py
@@ -10,6 +10,7 @@ class AccountAsset(models.Model):
     operating_unit_id = fields.Many2one(
         comodel_name="operating.unit",
         string="Operating Unit",
+        check_company=True,
         default=lambda self: self.env["res.users"].operating_unit_default_get(
             self.env.user.id
         ),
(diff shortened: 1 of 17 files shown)


AaronHForgeFlow:
Too bad. It seems OU modules are not multicompany consistent :S

The changes are good to me, and I think we should merge this. Multicompany production instances may need to check their data before applying this change. I will wait for other contributors to give their opinion first.

I think having a commit per module is required in OCA, sorry for the pain.

Saran440:
@AaronHForgeFlow No problem. I think it’s the right thing to do.
However, should we start fixing this from version 18?

Because so far, no one has multi ou in multi company.

AaronHForgeFlow:
I don't use this in a multicompany environment and I won't be affected. In any case I think it is best to fix the bug in any version.

AaronHForgeFlow:
I checked and it seems to be already solved in v17. I did not test v16.

It is ok to merge it and check if needed in v16

PR #751, opened 2025-03-11
[15.0][FIX] account_operating_unit: Inter OU Balance entry several issues

AaronHForgeFlow:
Steps to reproduce.

**Issue1: Inter OU Balance Entry is considered invoice**

Choose various invoices from different operating units and create a group payment.

Use a bank journal of one of the operating unit to pay.

The system will create the Inter-OU balance journal entry with type in_invoice. That is not correct. Indeed an error is being raised:

multi_ou_payment

**Issue 2: Inter OU balance Entry is unbalanced when matching partial amounts**

Create an invoice for 13000$ for main operating unit
Create a payment for 38000$ for B2B operating unit
Try to apply the payment to the invoice:
image

The system will create a debit for 38000$ and a credit for 13000$ for the InterOU Balance entry Causing an error for not balanced JE.

The solution is to InterOU Balance entry only for the amount being reconciled, 13000$

### DIFF
--- a/account_operating_unit/models/account_move.py
+++ b/account_operating_unit/models/account_move.py
@@ -120,6 +120,7 @@ class AccountMoveLine(models.Model):
             "date": max(self.mapped("date")),
             "ref": "Inter OU Balancing",
             "company_id": journal.company_id.id,
+            "move_type": "entry",
         }
         return move_vals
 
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (approved):
code review

MiquelRForgeFlow (approved):

PR #755, opened 2025-03-11
[18.0][MIG] purchase_stock_operating_unit: Migration to 18.

BhaveshHeliconia:


kanisorn-kate (approved):
LGTM

purchase_stock_operating_unit/models/purchase.py
```diff
@@ -1,24 +1,63 @@
...
 
-from odoo import _, api, models
+from odoo import api, fields, models
 from odoo.exceptions import UserError
 
 
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
+    picking_type_id = fields.Many2one(
```
Saran440, line 11: Functional Test :+1: 

But I'm not sure. Should we use onchange instead compute with `picking_type_id` and `operating_unit_id`?

AaronHForgeFlow: I think onchange makes more sense. Maybe a constraint if there is inconsistency with the operating unit.

BhaveshHeliconia:
I have addressed your feedback @Saran440 and @AaronHForgeFlow. Could you please review?

1. Refactored `picking_type_id` and `operating_unit_id` from computed fields to regular fields with `onchange` methods.
2. Kept the Python constraint to enforce consistency between the Operating Unit and the Picking Type.
3. Updated the test suite to use `Form` objects so the onchange logic is properly triggered and verified during testing.

Saran440 (approved):
Code Review 👍

AaronHForgeFlow (approved):
Code review LGTM

PR #763, opened 2025-04-15
[14.0][FIX]operating_unit: show only user-readable operating units

PicchiSeba:
In a multi-company environment there might be cases where our user is in the `operating_unit.group_manager_operating_unit` group and is trying to see another user's form view.
If the other user has been assigned to some Operating Units owned by another company, opening the view results in an access error because we don't have access to the other company's data.

With this PR we handle this case, filtering out all those OU we cannot access


francesco-ooops (approved):
Functional ok!

AaronHForgeFlow (commented):
If the issue is the company, can we just check the company instead or at least using _check_access_rights function

PicchiSeba:
@AaronHForgeFlow that's what I tried to do initially, but I wasn't able to make it work

francesco-ooops:
@PicchiSeba can you rebase?

PR #766, opened 2025-05-14
[14.0][FIX] operating_unit: get default OU from correct field

HekkiMelody:
This commit also clarifies the usage of the two OU-related fields in the user, both for devs and for end users.

Will FW up to v17 if approved

### DIFF
--- a/operating_unit/models/res_users.py
+++ b/operating_unit/models/res_users.py
@@ -19,11 +19,11 @@ class ResUsers(models.Model):
             return user.default_operating_unit_id
         else:
             # find an OU of the main active company
-            for ou in user.assigned_operating_unit_ids:
+            for ou in user.operating_unit_ids:
                 if ou.sudo().company_id in self.env.company:
                     return ou
             # find an OU of any active company
-            for ou in user.assigned_operating_unit_ids:
+            for ou in user.operating_unit_ids:
                 if ou.sudo().company_id in self.env.companies:
                     return ou
         return False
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (commented):
I am ok with the fix, but unsure about the need of the new field, operating_unit_readonly

HekkiMelody:
Thank you for taking my PR into consideration!

I added it because otherwise the behavior for the user is

- See a bunch of OUs on the user who is also OU manager
- Remove some and save
- The just-removed OUs reappear again.

Which is pretty frustrating and hard to explain...

(I'll also fix the tests ASAP)

francesco-ooops (approved):
Functional ok!

@AaronHForgeFlow what do you think?

francesco-ooops:
@AaronHForgeFlow ping :)

francesco-ooops:
@AaronHForgeFlow we're using this in production and is working well

AaronHForgeFlow:
@francesco-ooops Sorry I have not had the proper time to review this change. It is quite a big one and it is done in v14 should I have to consider also the impact in the newer versions. If some other contributors give more opinions maybe this could be merge sooner.

francesco-ooops:
@AaronHForgeFlow shall we start from later versions and then backport?

It is quite useful as a fix, especially in multi-company environments.

AaronHForgeFlow:
> @AaronHForgeFlow shall we start from later versions and then backport?

No. Just create the forward port when this is merged.

AaronHForgeFlow (approved):
Tested and I think  it is a good improvement!

AaronHForgeFlow:
/ocabot merge minor

@aleuffre can you please do the forward ports?

PR #774, opened 2025-06-06
[16.0][FIX] operating_unit: get default OU from correct field

HekkiMelody:
FW: https://github.com/OCA/operating-unit/pull/766/

This commit also clarifies the usage of the two OU-related fields in the user, both for devs and for end users.


AaronHForgeFlow (approved):

AaronHForgeFlow:
/ocabot merge minor

Merging to keep branches even

Issue #784, opened 2025-07-15
[16.0] sale_operating_unit: Issue with operating units in e-Commerce


tarteo:
Origin: https://github.com/OCA/operating-unit/pull/600
I'm not sure if this should be a separate module like website_sale_operating_unit or it should be handled here

To reproduce:

1. Install website_sale;
2. create a product and publish it;
3. make sure the sales team is set to "Website" on the website (this is by default)
4. create a OU "Online Orders" or whatever and change the operating unit on the sales team to it
5. go to /shop and click "Add to Cart"
6. This error occurs:
7. Configuration error. The Operating Unit of the sales team must match with that of the quote/sales order.

<img width="2143" height="661" alt="Image" src="https://github.com/user-attachments/assets/9e26c2ce-f40b-4577-b40c-e1bc91a0df25" />

I fixed it by replacing it with compute and inverse but keeping the onchanges for ux

``` python
    operating_unit_id = fields.Many2one(
        comodel_name="operating.unit",
        string="Operating Unit",
        compute="_compute_operating_unit_id",
        default=_default_operating_unit,
        store=True,
        readonly=True,
        states={"draft": [("readonly", False)], "sent": [("readonly", False)]},
    )

    @api.onchange("team_id")
    @api.depends("team_id", "team_id.operating_unit_id")
    def _compute_operating_unit_id(self):
        for sale in self.filtered(lambda s: s.state in ("draft", "sent") and s.team_id.operating_unit_id):
            sale.operating_unit_id = sale.team_id.operating_unit_id

    @api.onchange("operating_unit_id")
    def _onchange_operating_unit_id(self):
        for sale in self.filtered(
            lambda s: s.state in ("draft", "sent")
            and s.operating_unit_id != s.team_id.operating_unit_id
        ):
            sale.team_id = self.env["crm.team"].search(
                [("operating_unit_id", "=", sale.operating_unit_id.id)], limit=1
            )
```

This will also need a pre_init_hook, because the root user is assigned to an operating unit, which causes the sales team to be assigned an operating unit as well when the sales_team_operating_unit module is installed.

https://github.com/OCA/operating-unit/blob/16.0/operating_unit/data/operating_unit_data.xml

https://github.com/OCA/operating-unit/blob/77d307addc884574aedd5010b9d7e117d827bfde/sales_team_operating_unit/models/crm_team.py#L13-L16

[...]

PR #785, opened 2025-07-15
[16.0] operating_unit: Remove assigning operating unit to root user

tarteo:
This (the assigning of an operating unit to the root user) has a downstream impact where the sales teams get assigned an operating unit (sales_team_operating_unit) and consequently also all sales (sale_operating_unit) when the respective modules are installed. IMO this module should not touch any existing records and let records have initial null as operating unit.

Also explained in https://github.com/OCA/operating-unit/issues/784

### DIFF
--- a/operating_unit/data/operating_unit_data.xml
+++ b/operating_unit/data/operating_unit_data.xml
@@ -6,10 +6,6 @@
         <field name="code">OU1</field>
         <field name="partner_id" ref="base.main_partner" />
     </record>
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
ok for me!

tarteo:
@AaronHForgeFlow can this be merged? We've encountered another issue because of this.

AaronHForgeFlow:
@tarteo please forward port it to newer versions when you have the time. Thank you.

PR #787, opened 2025-07-25
[14.0][FIX] operating_unit: allow access to user without all companies active

HekkiMelody:
Steps to reproduce:

1. Install `operating_unit` on a fresh DB
2. Create Company 2
3. Activate only Company 2
4. Enter Marc Demo's user 

Error: 
`Due to security restrictions, you are not allowed to access 'Operating Unit' (operating.unit) records.`

### DIFF
--- a/operating_unit/models/res_users.py
+++ b/operating_unit/models/res_users.py
@@ -74,21 +74,23 @@ class ResUsers(models.Model):
                 record.default_operating_unit_id = False
 
     @api.depends("groups_id", "assigned_operating_unit_ids")
+    @api.depends_context("allowed_company_ids")
     def _compute_operating_unit_ids(self):
+        if self.env.context.get("allowed_company_ids"):
+            dom = [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "in", self.env.context["allowed_company_ids"]),
+            ]
+        else:
+            dom = []
         for user in self:
             if user.has_group("operating_unit.group_manager_operating_unit"):
-                dom = []
-                if self.env.context.get("allowed_company_ids"):
-                    dom = [
-                        "|",
-                        ("company_id", "=", False),
-                        ("company_id", "in", self.env.context["allowed_company_ids"]),
-                    ]
-                else:
-                    dom = []
-                user.operating_unit_ids = self.env["operating.unit"].sudo().search(dom)
+                user.operating_unit_ids = self.env["operating.unit"].search(dom)
             else:
-                user.operating_unit_ids = user.assigned_operating_unit_ids
+                user.operating_unit_ids = (
+                    user.assigned_operating_unit_ids.filtered_domain(dom)
+                )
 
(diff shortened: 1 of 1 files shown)


francesco-ooops (approved):
Functional ok!

francesco-ooops:
@AaronHForgeFlow we attempted this in https://github.com/OCA/operating-unit/pull/763 and settled for this solution in the end, wdyt?

operating_unit/models/res_users.py
```diff
@@ -74,21 +74,23 @@ def _onchange_operating_unit_ids(self):
...
+    @api.depends_context("allowed_company_ids")
     def _compute_operating_unit_ids(self):
+        if self.env.context.get("allowed_company_ids"):
+            dom = [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "in", self.env.context["allowed_company_ids"]),
+            ]
+        else:
+            dom = []
```
AaronHForgeFlow, line 86: when is this happening? in a non multi-company environment?

HekkiMelody: I'm not entirely sure, I couldn't find an example of when `allowed_company_ids` was not set. Still, I preserved it since it was already there, and also as a form of defensive code, just in case.

operating_unit/models/res_users.py
```diff
@@ -74,21 +74,23 @@ def _onchange_operating_unit_ids(self):
...
-                if self.env.context.get("allowed_company_ids"):
-                    dom = [
-                        "|",
-                        ("company_id", "=", False),
-                        ("company_id", "in", self.env.context["allowed_company_ids"]),
-                    ]
-                else:
-                    dom = []
-                user.operating_unit_ids = self.env["operating.unit"].sudo().search(dom)
+                user.operating_unit_ids = self.env["operating.unit"].search(dom)
```
AaronHForgeFlow, line 89: Not sure if removing the sudo is ok. That could lead to disrupts for current users?

operating_unit/models/res_users.py
AaronHForgeFlow: it is a bit strange but as it was there before it is ok.

operating_unit/models/res_users.py
HekkiMelody: The function is already `compute_sudo=True`. I removed it because it seemed superfluous.

HekkiMelody: If you believe it's safer I could keep it, of course, it doesn't *hurt*

AaronHForgeFlow: No, compute_sudo() is fine :)

Will you do the forward-port if this gets merged?

francesco-ooops: @AaronHForgeFlow of course we will

AaronHForgeFlow (approved):
Code review LGTM

PR #788, opened 2025-07-29
[16.0][FIX] operating_unit: allow access to user without all companies active

Chionne27:
FWP #787

PR #791, opened 2025-07-29
[17.0][FIX] operating_unit: allow access to user without all companies active

Chionne27:
FWP #787

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,10 +1,10 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.29
+_commit: v1.33
 _src_path: gh:oca/oca-addons-repo-template
 additional_ruff_rules: []
 ci: GitHub
 convert_readme_fragments_to_markdown: true
-enable_checklog_odoo: false
+enable_checklog_odoo: true
 generate_requirements_txt: true
 github_check_license: true
 github_ci_extra_env: {}
@@ -17,7 +17,8 @@ odoo_test_flavor: Both
 odoo_version: 17.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- res_partner_operating_unit
 repo_description: 'TODO: add repo description.'
 repo_name: operating-unit
 repo_slug: operating-unit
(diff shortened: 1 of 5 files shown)


AaronHForgeFlow:
Can you check tests? BTW the FWP in v16 was not done because it is not needed?

Chionne27:
> Can you check tests? BTW the FWP in v16 was not done because it is not needed?
 
The FWP in v16 is done in the following PR
https://github.com/OCA/operating-unit/pull/788/files

HekkiMelody:
I have a possible fix for the tests, though it's ugly

```py
    @classmethod
    def setup_company_data(cls, company_name, chart_template=None, **kwargs):
        """
        HACK: Compatibility with res_partner_operating_unit
        Without it, user cannot setup a ChartOfAccount in setUpClass
        """
        rule = cls.env.ref("res_partner_operating_unit.ir_rule_res_partner_allowed_operating_units", raise_if_not_found=False)
        if rule:
            rule.active = False
        res = super().setup_company_data(company_name, chart_template=chart_template, **kwargs)
        if rule:
            rule.active = True
        return res
```

The problem is that this PR limits the apparent `operating_unit_ids` on the partner based on active companies. The setUpClass in `account` tries to use a specific user to setup the Chart of Accounts, but this user has no operating units, so they're not allowed to write on the company, which belongs to an OU.

AaronHForgeFlow:
If we just put res_partner_operating_unit in the rebel_modules list the issue would be gone I think, maybe you can try that.

HekkiMelody:
> If we just put res_partner_operating_unit in the rebel_modules list the issue would be gone I think, maybe you can try that.

I thought that would be even less desirable, because then we're testing without that module (whereas with my solution we're only deactivating a rule, just for the setUpClass.

But we can do that, for sure.

Thank you for your assistance as always @AaronHForgeFlow

AaronHForgeFlow:
@aleuffre  I don't like to exclude modules but there are some other occurrences in other repositories. Thank you.

AaronHForgeFlow:
Hi @Chionne27,  is this ready?

AaronHForgeFlow:
code LGTM @Chionne27 just please confirm this is ready and I will merge in order to keep branches aligned.

Chionne27:
@AaronHForgeFlow this is ready. You can merge

AaronHForgeFlow (approved):

AaronHForgeFlow:
Merging, to keep branches even asap:

/ocabot merge patch

PR #793, opened 2025-08-18
[17.0][ADD] account_accountant_operating_unit

BT-anieto:
This module allows users to perform the reconciliation for companies with the flag
*Operating Units are self-balanced* set.


AaronHForgeFlow:
Thank you for the proposal. Unfortunately this can't be merged in OCA. It depends on enterprise module.

BT-anieto:
@AaronHForgeFlow Oh!! You're right. I close the PR then :) Thanks for the quick answer

PR #797, opened 2025-08-26
[16.0][MIG] sale_stock_operating_unit

Chionne27:


AaronHForgeFlow:
Hi @Chionne27 why was it migrated from v14 instead of v15?

In any case there is an open migration here: https://github.com/OCA/operating-unit/pull/614/ it is a bit stale but maybe you can review it.

Chionne27:
> Hi @Chionne27 why was it migrated from v14 instead of v15?
> 
> In any case there is an open migration here: #614 it is a bit stale but maybe you can review it.

Hi @AaronHForgeFlow. You are correct. I wrongly made the migration from v14. Now is from 15.0

AaronHForgeFlow:
Something is wrong, because the last migration commit I see is from v12

Chionne27:
> Something is wrong, because the last migration commit I see is from v12
@AaronHForgeFlow I made the migration with format-patch considering v15. Probably i have to fixup the commit which specifies the migration from v12

Chionne27:
> Something is wrong, because the last migration commit I see is from v12

 @AaronHForgeFlow I made the migration with format-patch considering v15. Probably i have to fixup the commit which specifies the migration from v12

AaronHForgeFlow:
@Chionne27 sorry, I did not notice that it was the same module as in here: https://github.com/OCA/operating-unit/pull/614 I think that one is active so I prefer to continue on that one

sale_stock_operating_unit/models/sale_order.py
```diff
@@ -9,10 +9,16 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    user_id = fields.Many2one(
```
AaronHForgeFlow, line 12: is this a field re-definition?

sale_stock_operating_unit/models/sale_order.py
```diff
@@ -25,15 +31,15 @@ def _default_warehouse_id(self):
...
     warehouse_id = fields.Many2one(
         comodel_name="stock.warehouse", default=_default_warehouse_id
     )
 
-    @api.onchange("team_id")
-    def onchange_team_id(self):
-        res = super(SaleOrder, self).onchange_team_id()
+    @api.depends("team_id")
+    def _compute_team_id(self):
+        res = super(SaleOrder, self)._compute_team_id()
```
AaronHForgeFlow: I think you can use

res = super()._compute_team_id()

AaronHForgeFlow (commented):
Code looks almost good to me

alan196:
I have checked this PR and I lookt it Good, just fix the issues noted by @AaronHForgeFlow  and I'll review 👍

sale_stock_operating_unit/models/sale_order.py
Chionne27: No it shouldn't be. I only need to set a default because in v15 user_id is filled with the user_id 👍

sale_stock_operating_unit/models/sale_order.py
Chionne27: Yes is the same. Thank you

francesco-ooops:
@AaronHForgeFlow @alan196 good to go?

AaronHForgeFlow (approved):
code LGTM

francesco-ooops (approved):
Functional ok!

PR #813, opened 2025-11-19
[19.0][MIG] analytic_operating_unit: Migration to 19.0

jcoux:


analytic_operating_unit/static/src/components/analytic_distribution.esm.js
```diff
@@ -16,7 +21,7 @@ patch(AnalyticDistribution.prototype, {
                             [
                                 "operating_unit_ids",
                                 "in",
-                                this.props.record.data.operating_unit_id[0],
+                                this.props.record.data.operating_unit_id.id,
```
yankinmax, line 24: Considering the code is being rewritten in the runtime, I would make this change safer.
Yep, OWL is used more and more and we should be able to access the `id` of an `operating_unit_id` object.
But, there is still some legacy code, I think, so I would go for:
```
const ou = this.props.record.data.operating_unit_id;
const ouId = ou?.id || ou?.[0];
```

yankinmax (approved):

analytic_operating_unit/static/src/components/analytic_distribution.esm.js
jcoux: When testing, the value of `this.props.record.data.operating_unit_id[0]` was now undefined in version 19.0.
That's why I replaced it by `.id`.

yankinmax: Ah ok. So, we don't need safety here, it even forces an error.
Thanks for explaining.

AaronHForgeFlow (approved):
Code review LGTM

vvrossem (approved):
LG!

> * If you are comfortable with squashing commits in Git, you could consider to squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

jcoux:
> LG!
> 
> > * If you are comfortable with squashing commits in Git, you could consider to squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

Done! :pray:

PR #820, opened 2025-12-01
[14.0][FIX] purchase_operating_unit: create PO with warehouse without OU

HekkiMelody:
This commit fixes an issue where, if there is only one warehouse visible to the user and that warehouse has no Operating Unit (so it's available for all regardless of OU), PO creation fails with an error.

In case of multiple warehouses being available (with and without OUs) the one with the OU is preferred and used over the others.

### DIFF
--- a/purchase_operating_unit/models/purchase_order.py
+++ b/purchase_operating_unit/models/purchase_order.py
@@ -101,11 +101,18 @@ class PurchaseOrder(models.Model):
             types = type_obj.search(
                 [
                     ("code", "=", "incoming"),
-                    ("warehouse_id.operating_unit_id", "=", self.operating_unit_id.id),
-                ]
+                    (
+                        "warehouse_id.operating_unit_id",
+                        "in",
+                        [self.operating_unit_id.id, False],
+                    ),
+                ],
             )
             if types:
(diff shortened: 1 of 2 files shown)


francesco-ooops (approved):
Functional ok!

AaronHForgeFlow:
@aleuffre  Typically there is no warehouse without operating unit in the company and I think the OU should be mandatory in the warehouse. But being the code in stock_operating_unit as it is, I am ok with this improvement. Would you do the forward port when merged?

HekkiMelody:
@AaronHForgeFlow Thank you for reviewing this. Of course I will do the forward ports.

AaronHForgeFlow (approved):

HekkiMelody:
I checked, and the `onchange` method I'm changing here is not present in higher versions of the module.

I'll do a quick functional check to see if the problem exists in higher versions, but it may be that forward porting is not even needed.

HekkiMelody:
Yes, `stock` is not even a dependency of the module in higher versions, so this change can't be forward-ported.

AaronHForgeFlow:
ok, thank you for checking. There is a purchase_stock_operating_unit module, but looking at the code it seems it does not have this issue.

PR #821, opened 2025-12-04
[FIX] Grant sudo to safely handle AccessError for the users without access to default operating unit

yankinmax:
The issue was originally raised while testing other modules even not depending on `operating_unit`

### DIFF
--- a/operating_unit/models/res_users.py
+++ b/operating_unit/models/res_users.py
@@ -2,6 +2,7 @@
 # - Jordi Ballester Alomar
 # Copyright 2015-TODAY Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
 from odoo import api, fields, models
 
 
@@ -13,18 +14,20 @@ class ResUsers(models.Model):
     def operating_unit_default_get(self, uid2=False):
         if not uid2:
             uid2 = self.env.user.id
-        user = self.env["res.users"].browse(uid2)
+        # Grant sudo for the user to avoid access rights issues while checking
+        # access to operating units.
+        user = self.env["res.users"].sudo().browse(uid2)
         # check if the company of the default OU is active
-        if user.default_operating_unit_id.sudo().company_id in self.env.companies:
+        if user.default_operating_unit_id.company_id in self.env.companies:
             return user.default_operating_unit_id
         else:
             # find an OU of the main active company
             for ou in user.operating_unit_ids:
-                if ou.sudo().company_id in self.env.company:
+                if ou.company_id == self.env.company:
                     return ou
             # find an OU of any active company
             for ou in user.operating_unit_ids:
-                if ou.sudo().company_id in self.env.companies:
+                if ou.company_id in self.env.companies:
                     return ou
         return False
 


operating_unit/models/res_users.py
```diff
@@ -14,6 +19,16 @@ def operating_unit_default_get(self, uid2=False):
...
         user = self.env["res.users"].browse(uid2)
+        # Check user read access on operating units.
+        # For other modules' (that even don't depend on ``operating_unit``) tests,
+        # we can have an AccessError raised for users without access to OUs.
+        # Therefore, we catch this exception and safely ignore it.
+        try:
+            self.env["operating.unit"].with_user(user).check_access_rights("read")
+            self.env["operating.unit"].with_user(user).check_access_rule("read")
+        except AccessError:
+            _logger.debug("User '%s' has no read access on operating units.", user.name)
```
vvrossem: It might be better to log it as a warning, as debug log level might not be activated
```suggestion
            _logger.warning("User '%s' has no read access on operating units.", user.name)
```
Although the OCA CI might not be happy about it

vvrossem (approved):

operating_unit/models/res_users.py
```diff
@@ -14,6 +19,16 @@ def operating_unit_default_get(self, uid2=False):
         if not uid2:
             uid2 = self.env.user.id
         user = self.env["res.users"].browse(uid2)
```
SilvioC2C: I would play it safer and just add a `sudo()` in this line:
```suggestion
        user = self.env["res.users"].sudo().browse(uid2)
```
Then you won't need to manually check if the user has permissions on either the model or specific records, and you can drop the rest of the `sudo()`s used below.

vvrossem (approved):

vvrossem:
:information_source: commit message should be `[FIX] <module>: <message>`

operating_unit/models/res_users.py
```diff
@@ -2,8 +2,12 @@
 # - Jordi Ballester Alomar
 # Copyright 2015-TODAY Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+import logging
+
 from odoo import api, fields, models
 
+_logger = logging.getLogger(__name__)
```
SilvioC2C: I think you can drop the loggers now

operating_unit/models/res_users.py
```diff
@@ -13,18 +17,20 @@ class ResUsers(models.Model):
...
+        user = self.env["res.users"].sudo().browse(uid2)
         # check if the company of the default OU is active
-        if user.default_operating_unit_id.sudo().company_id in self.env.companies:
+        if user.default_operating_unit_id.company_id in self.env.companies:
             return user.default_operating_unit_id
         else:
             # find an OU of the main active company
             for ou in user.operating_unit_ids:
-                if ou.sudo().company_id in self.env.company:
+                if ou.company_id in self.env.company:
```
SilvioC2C: Since we're already updating this line, can you switch the `in` operator to a `==`? The second term is just a single record, not a set.

SilvioC2C (approved):
LGTM, 2 minor remarks

AaronHForgeFlow (approved):
code LGTM

AaronHForgeFlow:
/ocabot merge patch

Please propose the forward ports when you have the time. Thanks.

PR #826, opened 2026-01-13
[17.0][FIX] crm_operating_unit: warning in logs

vvrossem:
Supersedes https://github.com/OCA/operating-unit/pull/718

When crm_operating_unit is installed, warning messages are sent to the Odoo logs:

WARNING odoodb_test odoo.fields: Redundant default on crm.lead.operating_unit_id

This is caused by the setting of a default value on a related field.

We change the field to be computed, read-only with a compute method which implements the logic of the default value in case the team of the lead does not have an operating unit.

### DIFF
--- a/crm_operating_unit/models/crm_lead.py
+++ b/crm_operating_unit/models/crm_lead.py
@@ -7,16 +7,22 @@ from odoo import api, fields, models
 class CRMLead(models.Model):
     _inherit = "crm.lead"
 
-    @api.model
-    def _get_default_operating_unit(self):
-        team = self.env["crm.team"]._get_default_team_id()
-        if team.operating_unit_id:
-            return team.operating_unit_id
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (commented):
LGTM, I think we can merge it, just a small question in the api.depends.

vvrossem:
Actually, I think we can close this PR as https://github.com/OCA/operating-unit/commit/23e57e34b0b26e7339cbf1462012284f1a964a15 takes care of the issue

vvrossem:
closing as `default` has been removed in https://github.com/OCA/operating-unit/commit/23e57e34b0b26e7339cbf1462012284f1a964a15

AaronHForgeFlow:
ah yes, I see that now. Thanks.

PR #827, opened 2026-01-14
[18.0][FW] [17.0][IMP] account_operating_unit: account.move::_default_operating_unit_id

vvrossem:
`account.move` should default on the journal's operating unit (if set) instead of the user's operating unit


AaronHForgeFlow (approved):
Thanks

PR #831, opened 2026-02-23
[19.0][IMP] operating_unit: Add sheet to form view

jcoux:
Add a `<sheet>` container to the form view.
This simplifies view inheritance in dependent modules, specifically to allow easier addition of a chatter.

### DIFF
--- a/operating_unit/view/operating_unit_view.xml
+++ b/operating_unit/view/operating_unit_view.xml
@@ -5,13 +5,15 @@
         <field name="model">operating.unit</field>
         <field name="arch" type="xml">
             <form string="Operating Unit">
-                <group name="main_group">
-                    <field name="name" />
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (commented):
yes, it should be there. I wonder if all the dependent modules -all the others in the repo- would still work after this change, or if the display will be affected, have you checked?

jcoux:
> yes, it should be there. I wonder if all the dependent modules -all the others in the repo- would still work after this change, or if the display will be affected, have you checked?

I have checked all other modules in this repository for inheritance of `view_operating_unit_form`, and none of them use it (at least in v19.0). This change should be safe.

AaronHForgeFlow (approved):
👍  Thank you

PR #834, opened 2026-03-24
[18.0][IMP] operating_unit: Add linked users stat button and bulk unit synchronization

BhaveshHeliconia:
This pull request introduces two key improvements to the operating_unit module to enhance user management and visibility.

### Key Improvements:
- Users Stat Button on Operating Units: Provides quick access to all users linked to an OU (including managers) via a Stat Button on the form view.
- Bulk Operating Unit Synchronization: A new server action on the Users list view to automatically assign OUs based on the user's allowed companies.


AaronHForgeFlow:
the stat button could be something useful, the server action I think it is not a generic enough feature to be included in the core module.

BhaveshHeliconia:
@AaronHForgeFlow, thanks for the feedback. I've removed the bulk synchronization feature and kept only the Stat Button. Documentation and tests have been updated accordingly.

operating_unit/models/operating_unit.py
```diff
@@ -56,3 +56,33 @@ def create(self, vals_list):
     def write(self, vals):
         self.env.registry.clear_cache()
         return super().write(vals)
+
+    def button_open_linked_users(self) -> dict:
```
AaronHForgeFlow: Can you explain why you need all this code for an action? why don't simply create the action in xml and do smth like this (haven't tested the code, just an idea):

def button_open_linked_users(self):
    action = self.env.ref("operating_unit.action_res_users_operating_unit").read()[0]
    if len(self.user_ids) == 1:
        action.update({
            "view_mode": "form",
            "res_id": self.user_ids.id,
        })
    else:
        action["domain"] = [("id", "in", self.user_ids.ids)]
    return action

BhaveshHeliconia: Thanks for the suggestion. I've improved the method accordingly.

BhaveshHeliconia:
ping @AaronHForgeFlow

PR #836, opened 2026-04-08
[17.0][IMP] mail_operating_unit: support single operating_unit_ids and add OU mail server routing

yankinmax:
This PR extends `mail_operating_unit` in two ways.

First, it adds support for records using `operating_unit_ids` in addition to `operating_unit_id`. When a `mail.thread` record has exactly one operating unit in `operating_unit_ids` and this operating unit has an alias domain, the module now uses that alias domain for mail generation. This covers models such as `res.partner`, which store operating units on a many2many field and are used by reminder emails. The existing behavior for `operating_unit_id` remains unchanged.

Second, it adds optional outgoing mail server routing based on the operating unit. When enabled in `General Settings`, and when no outgoing mail server is explicitly defined on the email template, the module selects the mail server from the target record:

- from `operating_unit_id` when it has a configured mail server;
- from `operating_unit_ids` when all operating units share the same mail server.

This allows emails such as reminders or invoices to be routed automatically through the proper SMTP server for each operating unit, while preserving template priority and falling back to standard Odoo behavior when no unambiguous operating-unit configuration can be determined.

### DIFF
--- a/mail_operating_unit/models/mail_mail.py
+++ b/mail_operating_unit/models/mail_mail.py
@@ -1,0 +1,64 @@
+# Copyright 2026 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+import logging
+
+from odoo import fields, models
+
+_logger = logging.getLogger(__name__)
+
+
+class MailMail(models.Model):
+    _inherit = "mail.mail"
+
+    operating_unit_id = fields.Many2one("operating.unit", string="Operating Unit")
+
+    def _send(
+        self,
+        auto_commit=False,
+        raise_exception=False,
+        smtp_session=None,
+        alias_domain_id=False,
+    ):
+        # OVERRIDE: to assign an OU mail server when no explicit mail server is set.
+
+        # Template mail server keeps the highest priority,
+        # because it is already stored on ``mail.mail.mail_server_id``
+        # before this method is called.
+        no_mail_server_mails = self.filtered(
+            lambda mail: not mail.mail_server_id and mail.model and mail.res_id
+        )
+        for mail in no_mail_server_mails:
+            # Check if the related record actually exists,
+            # as the mail could be linked to a deleted record.
+            record = self.env[mail.model].browse(mail.res_id).exists()
+            if not record:
+                continue
+
+            if mail_server := record._mail_get_operating_unit_mail_server():
+                mail.write({"mail_server_id": mail_server.id})
+                # Log the information about the resolved mail server
+                # and the OU for each mail being sent.
+                _logger.debug(
+                    "mail_operating_unit: using operating unit mail server %s (%s) "
+                    "from OU [%s] for record %s(%s)",
+                    mail_server.display_name,
+                    mail_server.id,
+                    record._mail_get_operating_unit_label(),
+                    record._name,
+                    record.id,
+                )
+            else:
+                _logger.debug(
+                    "mail_operating_unit: no operating unit mail server resolved "
+                    "from OU [%s] for record %s(%s), falling back to default behavior",
+                    record._mail_get_operating_unit_label(),
+                    record._name,
+                    record.id,
+                )
+
+        return super()._send(
+            auto_commit=auto_commit,
+            raise_exception=raise_exception,
+            smtp_session=smtp_session,
+            alias_domain_id=alias_domain_id,
+        )
--- a/mail_operating_unit/models/mail_template.py
+++ b/mail_operating_unit/models/mail_template.py
@@ -17,7 +17,8 @@ class MailTemplate(models.Model):
         email_values=None,
         email_layout_xmlid=False,
     ):
-        # Include in the email values the alias domain ID of the current OU if any
+        # OVERRIDE: to include in the email values
+        # the alias domain ID of the current OU if any
         email_values = email_values or {}
         if self.operating_unit_id and self.operating_unit_id.alias_domain_id:
             email_values["record_alias_domain_id"] = (
--- a/mail_operating_unit/models/models.py
+++ b/mail_operating_unit/models/models.py
@@ -10,16 +10,26 @@ class BaseModel(models.AbstractModel):
     def _mail_get_operating_unit(self):
         """Retrieves the operating unit ID if it exists and is truthy.
 
-        This method checks if the instance has the attribute `operating_unit_id`
+        This method checks if the instance has the attribute
+        either `operating_unit_id` or `operating_unit_ids`
         and whether it holds a truthy value.
-        If both conditions are met, it returns the value of `operating_unit_id`.
+        If both conditions are met, in case of `operating_unit_id`
+        it returns the value of `operating_unit_id`.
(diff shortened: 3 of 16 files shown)


mail_operating_unit/models/mail_template.py
```diff
@@ -30,3 +35,255 @@ def send_mail(
...
+        )
+
+    def _get_operating_unit_mail_server(self, record):
+        """Return the operating unit outgoing mail server.
+
+        If the record has ``operating_unit_id`` and that operating unit has
+        a configured outgoing mail server, use it.
+        If the record has ``operating_unit_ids`` and all operating units
+        share the same outgoing mail server, use that server.
+        Otherwise return `False`.
```
ivantodorovich:
```suggestion
        Otherwise, try to use the one of the operating unit the current user is logged on.
```

In case records have different mail servers, perhaps we should use the one linked to the currently logged company/operating unit, by calling the method `res.users._get_default_operating_unit`.

Because, if a record has multiple `operating_unit_ids`, it just means it's available on all of them.. but the currently active company (and thus, active operating unit) is the one that matters IMO.

mail_operating_unit/models/mail_template.py
```diff
@@ -30,3 +35,255 @@ def send_mail(
...
+        """
+        record.ensure_one()
+        if "operating_unit_id" in record and record.operating_unit_id:
```
ivantodorovich: ⚠️ , check `record._fields` instead

mail_operating_unit/models/mail_template.py
```diff
@@ -30,3 +35,255 @@ def send_mail(
...
+    @api.returns("self", lambda value: value.ids)
+    def send_mail_batch(
+        self,
+        res_ids,
+        force_send=False,
+        raise_exception=False,
+        email_values=None,
+        email_layout_xmlid=False,
+    ):
+        # OVERRIDE: to allow routing outgoing mail server from the operating unit.
```
ivantodorovich: We should exhaust all options before going for a complete overwrite of this method. It will be a pain to maintain in the long run.

Have you checked how the `mail_outbound_static` module handles this? I think it may apply in this case, too: https://github.com/OCA/social/blob/17.0/mail_outbound_static/models/ir_mail_server.py#L112

At that stage, you may be able to grab a reference to the `record` linked to the message (perhaps through the `message['object_id']`), and derive the mail server from there.

mail_operating_unit/models/mail_template.py
```diff
@@ -30,3 +35,255 @@ def send_mail(
             email_values=email_values,
             email_layout_xmlid=email_layout_xmlid,
         )
+
+    def _is_operating_unit_use_mail_server_enabled(self):
```
ivantodorovich: IMHO this settings is useless.

It should always attempt to do this. 

Users can decide whether to enable/disable this feature by simply not configuring any `mail_server_id` in their operating units.

mail_operating_unit/models/mail_template.py
ivantodorovich: NOTE: There are other cases like this

mail_operating_unit/models/mail_template.py
yankinmax: This feature is requested in the BA specifications :smile:
I agree we can just decide by the `mail_server_id` set on the OU.

mail_operating_unit/models/mail_template.py
yankinmax: Yep, I know we should avoid that.
I didn't find any other way so far to change the `mail_server_id` before mail is actually sent.
I'll check your suggestion.
Thanks for the review.

mail_operating_unit/models/mail_template.py
yankinmax: Makes sense. It was originally designed to return False for the mail alias domain, so I decided to follow the idea.

mail_operating_unit/models/models.py
```diff
@@ -72,3 +82,47 @@ def _mail_get_alias_domains(self, default_company=False):
...
+        - use ``operating_unit_id.mail_server_id`` when available;
+        - use the shared mail server from ``operating_unit_ids`` when all
+          operating units point to the same one;
+        - otherwise, if the current user's default operating unit belongs to
+          ``operating_unit_ids`` and has a mail server, use it;
+        - otherwise return ``False``.
+        """
+        self.ensure_one()
+        if "operating_unit_id" in self._fields and self.operating_unit_id:
+            return self.operating_unit_id.mail_server_id or False
```
vvrossem:
```suggestion
            return self.operating_unit_id.mail_server_id
```

mail_operating_unit/models/models.py
```diff
@@ -72,3 +82,47 @@ def _mail_get_alias_domains(self, default_company=False):
...
+        if "operating_unit_ids" in self._fields and self.operating_unit_ids:
+            mail_servers = self.operating_unit_ids.mapped("mail_server_id")
+            if len(mail_servers) == 1 and mail_servers:
```
vvrossem:
```suggestion
            if mail_servers and len(mail_servers) == 1:
```

mail_operating_unit/tests/common.py
```diff
@@ -1,28 +1,29 @@
...
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        # Email Alias Domain
+        # Email Alias Domains
         cls.ou_alias_domain = cls.env.ref("mail_operating_unit.ou_alias_domain")
-        cls.b2c_alias_domain = cls.env.ref("mail_operating_unit.b2b_alias_domain")
-        cls.b2b_alias_domain = cls.env.ref("mail_operating_unit.b2c_alias_domain")
+        cls.b2c_alias_domain = cls.env.ref("mail_operating_unit.b2c_alias_domain")
+        cls.b2b_alias_domain = cls.env.ref("mail_operating_unit.b2b_alias_domain")
```
vvrossem, line 19: nice catch! 🎯

vvrossem (approved):

mail_operating_unit/models/mail_mail.py
```diff
@@ -0,0 +1,64 @@
...
+                    record.id,
+                )
+            else:
+                _logger.info(
+                    "mail_operating_unit: no operating unit mail server resolved "
+                    "from OU [%s] for record %s(%s), falling back to default behavior",
+                    record._mail_get_operating_unit_label(),
+                    record._name,
+                    record.id,
+                )
```
ivantodorovich, line 57: Do we really want to keep these logs?
I think they could be quite annoying over time. Specially since the OU mail server is stored in the mail record anyway.. 
Perhaps keep only at the DEBUG level, so they can be explicitly enabled if needed

ivantodorovich (approved):

mail_operating_unit/models/mail_mail.py
yankinmax: Agree with you

yankinmax:
Hello @AaronHForgeFlow can you pls take a look and possibly trigger merge?

AaronHForgeFlow (approved):
👍 soft code review

PR #678, opened 2024-06-04
[17.0][MIG] `res_partner_operating_unit`: Migration to 17.0

yankinmax:
Based on:
- [ ] https://github.com/OCA/operating-unit/pull/397

Depends on:
- [x] https://github.com/OCA/operating-unit/pull/667

PR #837, opened 2026-04-08
[17.0] [FIX] res_partner_operating_unit: performance due to missing indexes

ivantodorovich:
There are missing indexes in a Many2many relation table because of this hook.

You see, the ORM takes care of creating indexes for relation tables, but only if its the one creating the table:
- https://github.com/odoo/odoo/blob/7d146774c/odoo/fields.py#L4804-L4817

The hook were manually creating the relation table and so bypassing the index creation, leading to performance issues (taking up to 15seconds to open the Contacts menu in our production db)

Looking into why the hook was created in the first place, I think it was wrongly included in one of the migrations.

You see, this code was originally proposed in `12.0`, here:
- https://github.com/OCA/operating-unit/pull/274

However, that PR was never merged. There was even a review explicitly stating the PR should not be merged:
- https://github.com/OCA/operating-unit/pull/274#issuecomment-845926729

Nonetheless, the commit was included in the migration to 14.0 -- which was never merged anyway:
- https://github.com/OCA/operating-unit/pull/397

Later, though, a migration to 17.0 proposal was made based on the mentioned migration to 14.0 work
- https://github.com/OCA/operating-unit/pull/678

---

As for the commit itself, I have to say I find no reason to make the `operating_unit_ids` field required, nor to hack the `ir.rule`. Maybe it was needed in `12.0` for some reason.. but not now.

The proposed solution is:
- Remove the `required=True` in the `res.partner.operating_unit_ids` field. The existing `ir.rule` already considers unset values.
- Remove the `pre_init_hook` that manually.
- For existing deployments, the migration script takes care of creating the missing indexes


vvrossem (approved):
LG!

SilvioC2C (approved):
LGTM

AaronHForgeFlow (approved):
Fair enough

ivantodorovich:
@AaronHForgeFlow is it ok to merge ? 🙏🏻

yankinmax:
Hello @OCA/accounting-maintainers can you pls take a look and possibly trigger merge?

PR #841, opened 2026-04-08
[17.0] [IMP] stock_operating_unit: add index

ivantodorovich:


vvrossem (approved):
LGTM!

SilvioC2C (approved):

AaronHForgeFlow (commented):
Is it really necessary in locations and warehouses? there are not that many

ivantodorovich:
For warehouses it may be too much, indeed.. but I figure it doesn't hurt neither.
For locations, you really can have quite a lot of them with nested locations.. so I think we do

AaronHForgeFlow:
Ok, I think then we should not add it to the warehouse, as it does not really help.

ivantodorovich:
Done ✅

AaronHForgeFlow (approved):
Thanks

AaronHForgeFlow:
/ocabot merge minor

:pray: the forward port when you have the time

PR #824, opened 2025-12-31
[14.0][FIX] account_operating_unit: Manager can select all OUs

SirPyTech:
Steps:
1. Create a Manager of Operating Units
2. Create some Operating Units
3. Log in as the created Manager
4. Create an Invoice
5. Click on the Operating Unit field

Before this change:
Not all Operating Units can be selected.

After this change:
All Operating Units can be selected.

PR #849, opened 2026-04-27
[14.0][IMP] stock_operating_unit_access_all: add stock views for all operating units access

GSLabIt:


francesco-ooops (approved):
Functional ok!

HekkiMelody (changes requested):
1. it's usually best to avoid using `replace` unless absolutely necessary. In this case, it'd be more compatible to simply operate on the `domain` attribute, I think. All other field customizations are lost otherwise.
2. The issue being fixed here is that a user with "Manage all OUs" doesn't seems to be able to select any OUs.
3.  Checking the field `user_ids` to limit the visible operating units is obsolete and unnecessary because the available operating units are already filtered by record rules automatically based on what the user has available. Additionally people who have all operating units available don't appear in the `user_ids` field, which is what causes the underlying issue. (This was already raised in [this PR](https://github.com/OCA/operating-unit/pull/766/).) So I think these changes shouldn't go here, but rather in `stock_operating_unit`. See [this PR](https://github.com/OCA/operating-unit/pull/824) for an example of the same fix. Most modules in `operating-unit` could potentially benefit from the same fix.

HekkiMelody:
Can you check if [this PR](https://github.com/OCA/operating-unit/pull/850) fixes the issue?

GSLabIt:
> 1. it's usually best to avoid using `replace` unless absolutely necessary. In this case, it'd be more compatible to simply operate on the `domain` attribute, I think. All other field customizations are lost otherwise.
> 2. The issue being fixed here is that a user with "Manage all OUs" doesn't seems to be able to select any OUs.
> 3. Checking the field `user_ids` to limit the visible operating units is obsolete and unnecessary because the available operating units are already filtered by record rules automatically based on what the user has available. Additionally people who have all operating units available don't appear in the `user_ids` field, which is what causes the underlying issue. (This was already raised in [this PR](https://github.com/OCA/operating-unit/pull/766/).) So I think these changes shouldn't go here, but rather in `stock_operating_unit`. See [this PR](https://github.com/OCA/operating-unit/pull/824) for an example of the same fix. Most modules in `operating-unit` could potentially benefit from the same fix.

Thanks for the review, you has been faster than me to apply changes. Closing this one in favor of #850

PR #856, opened 2026-05-21
[18.0][FIX] sales_team_operating_unit: Add default company_id

solomonprabu:
Adding default company to resolve incompatible records error coming up while creating sales team, when a user doesn't have multi-company access.

> [!WARNING]
>  image

### DIFF
--- a/sales_team_operating_unit/models/crm_team.py
+++ b/sales_team_operating_unit/models/crm_team.py
@@ -2,7 +2,7 @@
 # Copyright 2017-TODAY Serpent Consulting Services Pvt. Ltd.
 #   (<http://www.serpentcs.com>)
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class CrmTeam(models.Model):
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):

AaronHForgeFlow (commented):
Why don't you modify the _get_default_operating_unit function instead?

or Have you considered an onchange on operating_unit_id?

solomonprabu:
> Why don't you modify the _get_default_operating_unit function instead?
> 
> or Have you considered an onchange on operating_unit_id?

Thank you for the suggestions. I evaluated both approaches also a third approach, but decided on overriding `default_get` for the following reasons:

1. Modifying `_get_default_operating_unit`: While this would inject the default value, altering a method explicitly named for operating units to compute or handle base company logic breaks the **single-responsibility principle** and reduces **code readability**. Overriding `default_get` keeps the initialization logic clean and centralized.
2. Using an onchange method: Since the operating unit domain is already restricted to the user's current company, an onchange would introduce unnecessary RPC overhead and redundant computations during UI interactions. Also if the user decides to go with the default OU, the onchange method is left untriggered. All these makes it a less desirable approach.
3. Alternative (default attribute): Ideally, defining a default attribute directly on the `company_id` field would be the cleanest solution. However, since `company_id` is inherited from the Odoo core `crm.team` model, we cannot safely modify its attribute directly without risk of overriding core behavior or creating inheritance conflicts.

IMHO Overriding `default_get` is the most robust and standard Odoo approach here, especially since `company_id` remains invisible for users **without multi-company privileges** (`base.group_multi_company`), meaning the field relies entirely on backend initialization.

AaronHForgeFlow (approved):
Ok, thanks for the explanation. Please leave one commit only

BhaveshHeliconia (approved):

AaronHForgeFlow:
can you please FW port to 19?

PR #860, opened 2026-06-24
[17.0][FIX] mail_operating_unit: assign OU mail server before SMTP session grouping

jcoux:
When multiple outgoing mail servers coexist, ``mail.mail.send()`` calls
``_split_by_mail_configuration()`` to group mails and open SMTP sessions
before delegating to ``_send()``.

At grouping time, ``mail_server_id`` is NULL on records whose server is
resolved from an operating unit, so ``_find_mail_server()`` picks a server
based on priority and from-filter rules, potentially selecting the wrong
one.

The existing override in ``_send()`` writes the correct server to the
database, but by then the SMTP session is already established on the wrong
server.

Extract the OU mail server resolution into a dedicated method
``_resolve_operating_unit_mail_server()`` and call it from a new
``_split_by_mail_configuration()`` override, so ``mail_server_id`` is set
before grouping occurs and the SMTP session is opened on the correct server
from the start.

The ``_send()`` override is kept as a safety net for direct calls that
bypass ``_split_by_mail_configuration()``.

Add ``test_smtp_server_selection.py`` which uses ``MockSmtplibCase`` to
assert which server is actually passed to ``connect()``, covering both the
OU case (OU server must win over a higher-priority catch-all) and the no-OU
case (standard server selection must still apply).

### DIFF
--- a/mail_operating_unit/models/mail_mail.py
+++ b/mail_operating_unit/models/mail_mail.py
@@ -12,28 +12,21 @@ class MailMail(models.Model):
 
     operating_unit_id = fields.Many2one("operating.unit", string="Operating Unit")
 
-    def _send(
-        self,
-        auto_commit=False,
-        raise_exception=False,
-        smtp_session=None,
-        alias_domain_id=False,
-    ):
-        # OVERRIDE: to assign an OU mail server when no explicit mail server is set.
+    def _resolve_operating_unit_mail_server(self):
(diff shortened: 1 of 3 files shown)


vvrossem (approved):
LGTM.

yankinmax (approved):

BhaveshHeliconia (approved):
LGTM!

AaronHForgeFlow (approved):

PR #863, opened 2026-07-21
[16.0][MIG] stock_account_operating_unit: Migration to 16.0

eldhose-beo:
This PR migrates stock_account_operating_unit to 16.0, rebased on the real upstream 15.0 history (previous version was based on 13.0 and missed the 14.0/15.0 migration commits).

Ref: ADOO-2791

Also fixed along the way:

_generate_valuation_lines_data had a dead-code bug that silently prevented operating_unit_id from ever being stamped on stock valuation lines — fixed.
test_pickings was disabled (which is why the above went uncaught) — reactivated.
Dropped the deprecated force_company context key in favor of with_company(...).


eldhose-beo:
Codecov checks are failing because this is a migration PR and no functional changes or new tests were introduced.

shabeebcsbeo:
Hi @kittiu @max3903 @Saran440 

Could you please review this PR and share your feedback?
 If everything looks good, could you please take the necessary action to get this PR merged?

Thanks!

AaronHForgeFlow (changes requested):
Please make sure there are not commits after the migration to 16 commit, and the migration to 16.0 commit must be unique (aside from pre-commit fixes)

PR #864, opened 2026-07-24
[FIX] sale_operating_unit_sequence: resolve operating unit from its d…

baguenth:
…efault when absent from create vals

operating_unit_id is hidden in the view for users without the Multiple Operating Units permission, so the client never submits it and it is absent from vals on create. The same is true for any programmatic/API create that omits it. _update_sale_order_name only looked at vals.get("operating_unit_id"), so in both cases it silently did nothing and the operating unit's sale sequence was never applied, even though the record still received that operating unit via its field default a moment later (in base create(), which merges defaults after this hook runs).

Add _get_sequence_operating_unit_id(), which falls back to self.default_get(["operating_unit_id"]) when vals has no operating unit, so the sequence is generated consistently for UI and backend/API creates.

### DIFF
--- a/sale_operating_unit_sequence/models/sale_order.py
+++ b/sale_operating_unit_sequence/models/sale_order.py
@@ -15,9 +15,23 @@ class SaleOrder(models.Model):
         return super().create(vals)
 
     @api.model
+    def _get_sequence_operating_unit_id(self, vals):
+        """Resolve the operating unit to use for sequence generation,
+        falling back to the field's default when it is missing from vals."""
+        if "operating_unit_id" in vals:
+            # Key present, even if falsy: the caller explicitly requested no
+            # operating unit, so do not override it with the default.
+            operating_unit_id = vals["operating_unit_id"]
+        else:
+            operating_unit_id = self.default_get(["operating_unit_id"]).get(
+                "operating_unit_id"
+            )
+        return operating_unit_id
+
+    @api.model
     def _update_sale_order_name(self, vals):
         """Update sale order name in vals if operating unit has a sequence."""
-        operating_unit_id = vals.get("operating_unit_id")
+        operating_unit_id = self._get_sequence_operating_unit_id(vals)
         if operating_unit_id:
             ou_id = self.env["operating.unit"].browse(operating_unit_id)
             name = ou_id._get_next_sale_order_number()
--- a/sale_operating_unit_sequence/tests/test_sale_operating_unit_sequence.py
+++ b/sale_operating_unit_sequence/tests/test_sale_operating_unit_sequence.py
@@ -20,3 +20,104 @@ class TestSaleOperatingUnitSequence(TransactionCase):
         so2 = so1.copy()
         so2_sequence = so2.name
         self.assertNotEqual(so1_sequence, so2_sequence, "Sequences are different")
+
+    def test_create_sequence_operating_unit_id_absent_from_vals(self):
+        """The OU's sale sequence must apply even when operating_unit_id is
+        absent from create() vals - e.g. because the field is hidden for a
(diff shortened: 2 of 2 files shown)


baguenth:
Hey @AaronHForgeFlow ,

kindly asking for a review :)

sebinthomasbeo (approved):

sebinthomasst (approved):

parvathyvijayanbeo (approved):

sale_operating_unit_sequence/models/sale_order.py
```diff
@@ -14,10 +14,31 @@ def create(self, vals):
...
+        still be applied to the record, but only later, inside the base
+        ``create()``, i.e. after this naming logic already ran. Resolve that
+        same default here so the sequence is applied consistently regardless
+        of how the order was created.
+        """
+        operating_unit_id = vals.get("operating_unit_id")
+        if not operating_unit_id:
+            operating_unit_id = self.default_get(["operating_unit_id"]).get(
+                "operating_unit_id"
+            )
```
AaronHForgeFlow: Ok to cover the case when OU is explicitly False?

```suggestion
        if "operating_unit_id" in vals:
            operating_unit_id = vals["operating_unit_id"]
        else:
            operating_unit_id = self.default_get(["operating_unit_id"]).get(
                "operating_unit_id"
            )
```

sale_operating_unit_sequence/models/sale_order.py
```diff
@@ -14,10 +14,31 @@ def create(self, vals):
             vals = self._update_sale_order_name(vals)
         return super().create(vals)
 
+    @api.model
+    def _get_sequence_operating_unit_id(self, vals):
+        """Resolve the operating unit to use for sequence generation.
```
AaronHForgeFlow: can you summarize this a bit? it is too much explanation for so little code

sale_operating_unit_sequence/tests/test_sale_operating_unit_sequence.py
```diff
@@ -20,3 +20,59 @@ def test_create_sequence(self):
...
+                ],
+                "assigned_operating_unit_ids": [(6, 0, [operating_unit.id])],
+                "default_operating_unit_id": operating_unit.id,
+            }
+        )
+        self.assertFalse(
+            user.has_group("operating_unit.group_multi_operating_unit"),
+            "Test user must not have the Multiple Operating Units permission",
+        )
+
```
AaronHForgeFlow: avoid empty lines within methods

sale_operating_unit_sequence/tests/test_sale_operating_unit_sequence.py
```diff
@@ -20,3 +20,59 @@ def test_create_sequence(self):
...
+                {
+                    "partner_id": self.customer.id,
+                    # Avoid interference from the unrelated team/OU
+                    # consistency constraint (sale_operating_unit): this test
+                    # is only about the sequence/OU default resolution.
+                    "team_id": False,
+                }
+            )
+        )
+
```
AaronHForgeFlow: same as above

AaronHForgeFlow (commented):
LGTM just some suggestions

sale_operating_unit_sequence/models/sale_order.py
baguenth: > Ok to cover the case when OU is explicitly False?

Important point actually. Good catch :)

baguenth:
> LGTM just some suggestions

Thank you for the review :) Updated the code as per your suggestions.

baguenth:
@AaronHForgeFlow Can we integrate?

AaronHForgeFlow (approved):

